PR #1, opened 2024-12-09
#339208 https for dorc

kirill-suspitsin-ext-sefe:


src/Dorc.Api/Program.cs
```diff
@@ -91,6 +95,19 @@
 builder.Services.AddSingleton<IRateLimitConfiguration, RateLimitConfiguration>();
 builder.Services.AddInMemoryRateLimiting();
 
+//builder.Services.AddHttpsRedirection(options =>
```
andrei-salanovich-ext-sefe, line 98: let's remove redundant comments

kirill-suspitsin-ext-sefe: I suggest not to delete these comments. We will force https-only dorc mode a bit later and this is the tested working code for this goal

andrei-salanovich-ext-sefe: In this case I'd suggest to make feature flag for that instead keeping 'force https' functionality in comments, but feel free to leave as is.

src/Dorc.Api/Program.cs
```diff
@@ -102,6 +119,7 @@
 app.UseMiddleware<OptionsMiddleware>();
 app.UseCors(dorcCorsRefDataPolicy);
 
+//app.UseHsts();
```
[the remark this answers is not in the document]
andrei-salanovich-ext-sefe: In this case I'd suggest to make feature flag for that instead keeping 'force https' functionality in comments, but feel free to leave as is.

PR #21, opened 2025-01-31
#355037 export passwords for the secure variables available for the P…

kirill-suspitsin-ext-sefe:
…owerUsers


kirill-suspitsin-ext-sefe:
After discussion we decided to implement this functionality another way, PBI was closed

PR #22, opened 2025-01-31
Fix for the move of properties to configvalues

kirill-suspitsin-ext-sefe:
remove hashes

change sql test


src/Tools.DeployCopyEnvBuildCLI/Program.cs
```diff
@@ -24,27 +24,16 @@ private static int Main(string[] args)
...
-                var whiteListProperty = propertiesPersistentDataSource.GetGlobalProperties()
-                    .ToList()
-                    .FirstOrDefault(p=>p.Property.Name.Equals("DORC_CopyEnvBuildTargetWhitelist",StringComparison.CurrentCultureIgnoreCase));
-                if (whiteListProperty != null) strTargetEnvWhiteList = whiteListProperty.Value;
-            }
-            catch
-            {
-                Output("Couldn't find property DORC_CopyEnvBuildTargetWhitelist");
-            }
+            var whiteList = configValuesPersistentSource.GetConfigValue("DORC_CopyEnvBuildTargetWhitelist");
```
andrei-salanovich-ext-sefe, line 30: it can return null, should not we write to Output in that case?

src/Dorc.Monitor/ScriptDispatcher.cs
```diff
@@ -271,14 +271,9 @@ private string GetDeploymentRunnerFileFullName(
...
-        private string GetPropertyValue(string propertyName, string environmentName)
+        private string GetConfigValue(string propertyName)
         {
-            var propertyValues = propertyValuesPersistentSource.GetPropertyValues(propertyName, environmentName, true);
-
-            if (propertyValues.Any())
-                return propertyValues.First().Value.ToString();
-            throw new ApplicationException(
-                $"Unable to locate a value for the property {propertyName} for Environment {environmentName}");
+            return _configValuesPersistentSource.GetConfigValue(propertyName);
```
andrei-salanovich-ext-sefe, line 276: looks like it's better to throw exception here like it was before in the case config value does not exist, because otherwise we'll get some null reference exception somewhere higher up the call stack

hegsie (approved):

benhegartysefe (approved):

benhegartysefe (approved):

src/Tools.DeployCopyEnvBuildCLI/Program.cs
```diff
@@ -24,27 +24,19 @@ private static int Main(string[] args)
             //Console.WriteLine(container.WhatDoIHave());
 
             var deployLibrary = container.GetInstance<IDeployLibrary>();
-            var propertiesPersistentDataSource = container.GetInstance<IPropertyValuesPersistentSource>();
+            var configValuesPersistentSource = container.GetInstance<IConfigValuesPersistentSource>();
             var intReturnCode = 0;
             
-            var strTargetEnvWhiteList="";
-            try
+            var whiteList = configValuesPersistentSource.GetConfigValue("DORC_CopyEnvBuildTargetWhitelist");
```
andrei-salanovich-ext-sefe, line 30: maybe move "DORC_CopyEnvBuildTargetWhitelist" to some variable?

src/Dorc.Monitor/ScriptDispatcher.cs
```diff
@@ -271,16 +271,14 @@ private string GetDeploymentRunnerFileFullName(
...
+        private string GetConfigValue(string configValue)
         {
-            var propertyValues = propertyValuesPersistentSource.GetPropertyValues(propertyName, environmentName, true);
-
-            if (propertyValues.Any())
-                return propertyValues.First().Value.ToString();
-            throw new ApplicationException(
-                $"Unable to locate a value for the property {propertyName} for Environment {environmentName}");
+            if (string.IsNullOrEmpty(configValue))
+                throw new ApplicationException($"Config value name is empty, should have a value");
```
andrei-salanovich-ext-sefe, line 277: it would be good to know also which config value is empty

kirill-suspitsin-ext-sefe: We are just checking the method params on not empty value. We are not checking the value of the property from config here

src/Tools.DeployCopyEnvBuildCLI/Program.cs
kirill-suspitsin-ext-sefe: Fixed

PR #37, opened 2025-02-27
#358748 Fix an issue with changing property value

kirill-suspitsin-ext-sefe:


src/Dorc.Api/Services/PropertyValuesService.cs
```diff
@@ -32,6 +35,7 @@ public PropertyValuesService(ISecurityPrivilegesChecker securityPrivilegesChecke
             _propertiesPersistentSource = propertiesPersistentSource;
             _propertyEncryptor = propertyEncryptor;
             _securityPrivilegesChecker = securityPrivilegesChecker;
+            _log = LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);
```
andrei-salanovich-ext-sefe: let's inject that from constructor's parameter, ILog

andrei-salanovich-ext-sefe (approved):

PR #57, opened 2025-03-26
implement hybrid authentication - NTLM + Oauth

andrei-salanovich-ext-sefe:


src/Dorc.Core/Security/ConfigAuthSchemeEnum.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Dorc.Core.Security
+{
+    public static class ConfigAuthScheme
+    {
+        public const string OAuth = "OAuth";
+        public const string WinAuth = "WinAuth";
+        public const string Both = "Both";
```
atarasovSF, line 7: Please consider a more obvious constant name/value instead of `"Both"`. I suggest something like `"OAuth,WinAuth"` or `"OAuth+WinAuth"`

andrei-salanovich-ext-sefe: fixed

kirill-suspitsin-ext-sefe (approved):

PR #336, opened 2025-10-06
Fix/241 redesign add and attach buttons on db page of environment

slafeer-sefe:
got rid of add button and switched attach button from toggle to normal button


andrei-salanovich-ext-sefe:
this PR removes completely functionality of adding database from Environment->Databases page, is that what was meant in the tiсket?
image

Why do we need a button ATTAH DATABASE then at all? Seems it's a redundant click, we can hide attach dialog in 'Application Database Details'

dmitryrext:
hi @slafeer-sefe  let's do a rework - i've update the issue's description

PR #384, opened 2025-10-30
242 - redesign "add" and "attach" buttons on "servers" page of environment

slafeer-sefe:
removed add switch and changed attach button functionality to open dialog instead of dropdown


andrei-salanovich-ext-sefe:
it would be nice to see this in action on some environment like ST

andrei-salanovich-ext-sefe (approved):

PR #412, opened 2025-11-14
Revert "added new search filtering to allow for searching by instance aswell"

slafeer-sefe:
Reverts sefe/dorc#396

### DIFF
--- a/src/dorc-web/src/components/attach-database.ts
+++ b/src/dorc-web/src/components/attach-database.ts
@@ -267,16 +267,4 @@ export class AttachDatabase extends LitElement {
     });
     console.log(result);
   }
-
-  private filterDatabases(e: CustomEvent) {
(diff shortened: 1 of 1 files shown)


andrei-salanovich-ext-sefe (approved):

src/dorc-web/src/components/attach-database.ts
```diff
@@ -267,16 +267,4 @@ export class AttachDatabase extends LitElement {
     });
     console.log(result);
   }
-
-  private filterDatabases(e: CustomEvent) {
```
benhegartysefe: why are we removing this? surely we need to just add more elements to it

slafeer-sefe: have the complete code in an old local branch need to make a pr with it again since it was deleted with a merge conflict

PR #435, opened 2025-11-27
389-non-selected-components-deployed-bug

slafeer-sefe:


andrei-salanovich-ext-sefe (changes requested):
copilot comments are valid, at least first two (about inconsistent logic and adding a test for that)

src/Dorc.Core.Tests/DeployLibraryTests.cs
```diff
@@ -0,0 +1,199 @@
...
+                IsEnabled = true,
+                Children = new List<ComponentApiModel> { disabledChild }
+            };
+
+            // Act
+            InvokeAddComponent(componentNames, parent);
+
+            // Assert
+            Assert.AreEqual(1, componentNames.Count, "Only parent should be included");
+            Assert.IsTrue(componentNames.Contains("EnabledParent"), "Parent should be included");
```
andrei-salanovich-ext-sefe: there is no case when we deploy parent. Only leaves in the component tree are deployable

src/Dorc.Core.Tests/DeployLibraryTests.cs
```diff
@@ -0,0 +1,199 @@
...
+                ScriptPath = "parent.ps1",
+                IsEnabled = true,
+                Children = new List<ComponentApiModel> { child }
+            };
+
+            // Act
+            InvokeAddComponent(componentNames, parent);
+
+            // Assert
+            Assert.AreEqual(4, componentNames.Count, "Should include parent, child, and two grandchildren");
```
andrei-salanovich-ext-sefe: I suppose here should be only 1 as only 1 grandchild is enabled

andrei-salanovich-ext-sefe (changes requested):
see comments. IsEnabled cannot be nullable

src/Dorc.Core.Tests/DeployLibraryTests.cs
slafeer-sefe: changed tests to account for this

src/Dorc.Core.Tests/DeployLibraryTests.cs
slafeer-sefe: changed

andrei-salanovich-ext-sefe (approved):

PR #454, opened 2025-12-30
411: fix clearing PSVersion and not clearing ScriptPath

andrei-salanovich-ext-sefe:


src/Dorc.PersistentData/Sources/ManageProjectsPersistentSource.cs
```diff
@@ -450,14 +451,10 @@ public void UpdateComponent(ComponentApiModel apiComponent, int projectId, int?
                             component.Script = script;
                         }
                     }
-                    else if (apiComponent.ScriptPath == null && oldScript != null)
+                    else if (oldScript != null) // new script path is empty, removing script record
```
kirill-suspitsin-ext-sefe: New script path is not empty, it's null here.

andrei-salanovich-ext-sefe: fixed

PR #457, opened 2026-01-06
380-make-dorc-tools-use-api

### LINKED ISSUE #380: Make Dorc tools use API instead of shared code accessing DB directly
The tool Tools.DeployCopyEnvBuildCLI is using DeployLibrary class to create deployment requests. Since we want to have a notification when new requests are created, there are two options:
1) implement IDeploymentEventsPublisher similar to SignalRDeploymentEventPublisher from Monitor (or reuse it), configure access to Dorc API to send events
2) add API specially for the tool which will handle all events internally
3) CLi should also use DIS to authenticate to DORC API

Suggest to implement option #2 not only for mentioned tool but also migrate other tools to use Dorc API instead of shared code.
tools list:
ApplyPermissions\
ApplyPermissionsArc\
ApplyPermissionsComOps\
ApplyPermissionsComOps_backup\
ApplyPermissionsEndur\
DeployCopyEnvBuild\
DropDatabase\
EncryptionMigration\
PostRestoreEndur\
PostRestoreEndur_failing\
PropertyValueCreation\
RecModelShrink\
Request\
Request_backup\
Request_NTLM\
SQLBackup\
SQLRestore\

### LINKED ISSUE #546: Update MLP restrictions
Currently the tool "Tools.DeployCopyEnvBuildCLI.exe" can be used only for the environments that were added to DOrc config variable DORC_CopyEnvBuildTargetWhitelist
The other envs get error like"### is not a supported target env", e.g. https://dev.azure.com/sefe/DevOps/_workitems/edit/404297

Currently there are 46 envs in DORC_CopyEnvBuildTargetWhitelist
This restriction doesn't seem to be necessary

Please consider the possibility to lift the restriction or to provide the opportunity to add the envs to the white list for all Dorc users 
UPD. requirements were updated - please see comment https://github.com/sefe/dorc/issues/546#issuecomment-4030558022

slafeer-sefe:
Moved OauthConfig to Dorc.Core and made copyenvbuild tool use api


benhegartysefe:

### 1. `ConfigValuesController.cs` is at the project root, not in `Controllers/`
**Severity:** LOW • **Category:** architecture • **Location:** `src/Dorc.Api/ConfigValuesController.cs`

All other controllers in the project live under `src/Dorc.Api/Controllers/`, but this new file is added at `src/Dorc.Api/ConfigValuesController.cs` while still declaring `namespace Dorc.Api.Controllers`. Routing still works (ASP.NET discovers by type), but the structural inconsistency breaks the project's folder-equals-namespace convention and makes the file harder to discover alongside its siblings — exactly the kind of cohesion drift CLAUDE.md cautions against.

**Suggested fix:** Move the file to `src/Dorc.Api/Controllers/ConfigValuesController.cs` so it sits next to its peers.

---

**Other notable findings that fell short of strict 3-way consensus** (worth considering but flagged by only 2/3 reviewers):

- The new `GET /ConfigValues` endpoint has only `[Authorize]` and returns decrypted secure values via `IConfigValuesPersistentSource.GetConfigValue` — any authenticated user can read any secret stored in the table. (Opus + Sonnet, **HIGH**.)
- `Tools.DeployCopyEnvBuildCLI/Program.cs` calls `api.Call<EnvironmentApiModel>(Endpoints.RefDataEnvironments, ...)` but the endpoint always returns `List<EnvironmentApiModel>`. `JsonSerializer.Deserialize` throws, `IsModelValid=false`, every input is rejected as "not a valid target environment" — the CLI is non-functional for all inputs. (Opus + Sonnet, **HIGH**.)
- `DORC_CopyEnvBuildTargetWhitelist` enforcement was removed and replaced with only an `EnvironmentIsProd` flag on the API side — silently relaxes a pre-existing prod-target safety guard with no migration note. (Opus + Sonnet, **HIGH/MEDIUM**.)
- `RefDataEnvironmentsHistoryController.Post` writes caller-supplied `UpdatedBy` verbatim with no env permission check. (Opus + Sonnet.)
- `CopyEnvBuildController.Post` is `async Task<IActionResult>` but contains no `await` — produces CS1998, runs synchronously. (Opus + Haiku.)
- `RefreshEndur.UpdateEnvironmentHistory` swallows API-failure result silently, returning success. (Sonnet + Haiku.)

These are surfaced for awareness only — they did not meet the strict consensus bar.

---

kirill-suspitsin-ext-sefe (approved):

PR #516, opened 2026-02-25
Responsive layout, mobile-first design, and accessible drawer modal

### LINKED ISSUE #555: review mobile viewport version PR

hegsie:
## Summary

Responsive / mobile-first overhaul of the DOrc web UI: layout, navigation drawer, and the accessibility contract around them. Plus a year of `main` reconciled and a 3-round adversarial review.

### Core infrastructure

- **`ResponsiveMixin`** — reactive `_narrowScreen` driven by `matchMedia('(max-width: 768px)')`, exported `NARROW_BREAKPOINT` constant. Listeners attached/detached in `connectedCallback` / `disconnectedCallback`.
- **Viewport meta tag** + `100vh` / `100dvh` dual declarations so the mobile address bar doesn't eat layout.
- **`PageElement` / `PageLocation`** — alignment with `main`'s simplification (avoids TS6 deep-recursion on `RouterLocation<RouteMeta>`).

### Mobile drawer as an accessible modal

The hamburger ↔ drawer interaction was redesigned end-to-end:

- Drawer is `position: fixed` with a translate-X animation on narrow screens, occupies the document flow on desktop.
- **State machine** for `aria-*` and `inert`: static `role="navigation"` / `aria-label="Primary"` when closed or on desktop; temporarily upgraded to `role="dialog" aria-modal="true" aria-label="Navigation"` while modal-open on mobile.
- **`inert`** on `#page-content` while modal-open so screen-reader virtual cursors can't browse behind the drawer. `#header` is intentionally NOT inerted so the hamburger close button + Sign Out + Help stay reachable (locked in by test).
- **Escape key** closes the modal. **Router navigation** (`vaadin-router-location-changed`) auto-closes the drawer on mobile.
- **Focus** is moved into the drawer on open and restored to the previously-focused element on close, using a composed-path walk that crosses shadow boundaries.
- **`aria-controls="dorcNavbar"`** + `aria-expanded` on the hamburger button.
- **Ownership flags** (`_drawerLockedScroll`, `_drawerSetTabindex`) — we only release `body.style.overflow` and only strip `tabindex` if *we* set them, so coexisting modals/overlays aren't clobbered.
- **Splitter drag** (desktop) — listeners are idempotent across attach/detach cycles; `disconnectedCallback` only releases body styles it owns.

### Layout

- Replaced fixed-height patterns (`calc(100vh - Xpx)`) with `flex: 1; min-height: 0`. List pages now use `:host { display: flex; flex-direction: column; height: 100%; }` with the grid as the flex-growing child.
- Card components (`environment-card`, `project-card`, `page-project-envs`) converted from `position: absolute` to flexbox with `justify-content: space-between`.
- Cards stretch to `width: 100%` on narrow screens with a `max-width` constraint on wider screens.
- Table-based form layouts in `deploy-env` and `page-deploy` replaced with flex columns.

### Component / theme updates

- **Grid columns** — `ResponsiveMixin` applied to `page-databases-list`, `page-environments-list`, `page-servers-list`, `page-env-history`, `page-projects-list`, `page-monitor-requests`, etc. Secondary columns conditionally hidden on narrow screens.
- **Touch targets** — global `vaadin-button[theme~='icon']` floor of 44×44 px, **gated behind `@media (max-width: 768px)`** so compact desktop grids (`--lumo-button-size: 28px`) aren't blown up.
- **Dialog sizing** — global `width: min(90vw, 650px); max-height: 85vh/85dvh`. Opt-in `theme="log-viewer"` (Ace editor → 1400 px) and **opt-in `theme="wide"`** for the multi-column / diff dialogs: `terraform-plan-dialog`, `bundle-editor-dialog`, `make-like-production-dialog`.
- **Combo boxes** — `width: 100%; max-width: 600px` on narrow screens, scoped behind a media query so desktop sites that explicitly set `min-width: 490px` aren't overridden. The `add-edit-environment` form, which sets `min-width: 490px` in its own light-DOM scope, gets a local mobile override.
- **`hegs-chart`** — re-initialises the ECharts instance + `ResizeObserver` in `connectedCallback` when reconnected after `disconnectedCallback` disposed them (LitElement does not re-fire `firstUpdated`).
- **Lumo tokens** — replaced hardcoded font sizes / spacing with `var(--lumo-font-size-*)` and `var(--lumo-space-*)` throughout. Removed `padding: 2px` on buttons that was shrinking touch targets.

### Testing

- Migrated to **Vitest browser mode** (Playwright provider, multi-browser: chromium / firefox / webkit) and removed `web-test-runner`.
- Tests cover: `ResponsiveMixin` reactivity to MQ changes, drawer auto-close on router navigation + Escape + modal contract (`#page-content` inert / `#header` reachable), splitter listener survives reconnect, grid column hiding, card layout structure, touch-target media-query regression guard, dialog sizing.
- **95 tests / 10 files / all green** in chromium locally. CI matrix runs all three browsers.
- `tsconfig.test.json` for test-only TS config; `tsconfig.json` excludes `tests/**` AND `src/**/*.test.ts` (the second pattern protects against tests accidentally placed under `src/`).

### Build / CI

- Playwright cache + install in `release.yml` gated behind `runner.os == 'Windows'` so the matrix can be safely expanded later without silent CI mis-fires.
- `jsdom` devDep dropped (carried in from main; we use Playwright, not jsdom).

## Reconciliation with `main`

This branch was 65 commits behind `main` when reopened. Merging in resolved 8 file-level conflicts; resolution notes are in the merge commit `d38107c`. Notable choices:

- Kept main's GitHub-Actions-aware combo-box labels in `deploy-env.ts` (`isGitHubProject ? 'Workflow' : 'Build Definition'`); dropped `style="width: 600px"` (mobile responsive intent).
- Kept main's `_projectNameRenderer` on the project Name column AND the PR's new Description / Build-Regex columns.
- Preserved main's `"Artefacts Sub-Paths"` header on the ArtefactsSubPaths column (the PR had renamed it to "Azure DevOps Project(s)", which is no longer accurate now that main adds GitHub Actions build support).

## Adversarial review

Three rounds of consensus review with 3 reviewers per round (Opus 4.7, Sonnet 4.6, Haiku 4.5), each running independently. Round 5 closed 13 findings, round 6 closed 6, round 7 converged at **0 HIGH/MEDIUM findings** with all earlier fixes verified.

Highlights of what the panel caught:

- The global 44×44 touch-target floor was neutralising compact-grid icon buttons site-wide; now mobile-only.
- The original `focusin` guard for the modal drawer was actually preventing keyboard users from reaching header controls (Sign Out, Help); replaced with `inert` + scoped focus restoration.
- `_applyDrawerAria` was unconditionally releasing `document.body.style.overflow`, which could clobber a coexisting modal's body-lock.
- `hegs-chart` would silently render blank after detach/reattach because `firstUpdated` doesn't re-run.
- Playwright workflow steps were inconsistently gated for OS, a latent CI hazard if the matrix ever expanded.

Deferred (explicitly, with rationale in the round-5 and round-6 commit messages): chai/vitest `expect` stylistic mix, `:host { height: 100% }` parent-constraint assumption, responsive-mixin test mock permissiveness, dataProvider callback async timing weakness, hegs-chart `option` mutation between disconnect and reconnect (speculative).

## Verification

- `tsc --noEmit -p tsconfig.json && tsc --noEmit -p tsconfig.test.json` — clean
- `eslint src/**/*.ts tests/**/*.ts` — clean
- `vitest run` in chromium — 95 / 95 green


benhegartysefe:

This PR was reviewed independently by 3 agents (Claude Opus 4.7, Sonnet 4.6, Haiku 4.5). Findings where **all three** independently flagged the same issue area are posted below; strong 2/3 findings follow.

### 1. The mobile drawer auto-close handler `_drawerClickHandler` is buggy
**Severity:** MEDIUM • **Category:** correctness • **Location:** `src/dorc-web/src/components/dorc-app.ts` (`firstUpdated` → `_drawerClickHandler`, ~line 267–282)

All three reviewers independently flagged the auto-close handler — `e.composedPath().some(el => (el as HTMLElement).tagName === 'A')` — for distinct concrete failures:

- **Misses programmatic / router navigation** (Sonnet): `Vaadin Router` calls invoked via `router.go(...)` produce no anchor-click event in the path, so the drawer stays open after a programmatic navigation.
- **Misses keyboard activation** (Opus): keyboard-driven tab activation can dispatch synthetic clicks where the originating target is the tab (not the `<a>`), and the cross-shadow `composedPath` can omit the anchor depending on event composition.
- **Scope is too broad** (Haiku): the check fires on *any* anchor click anywhere in the event path — not just navbar links. A click on an `<a>` that opens a modal (or any other anchor in the shadow tree) will incorrectly close the drawer.

Together the handler is unreliable in both directions: it sometimes doesn't close when it should, and sometimes closes when it shouldn't.

**Suggested fix:** Listen for Vaadin Router's `vaadin-router-location-changed` event on `window` and close the drawer at that point. Decoupled, reliable across all navigation sources, and handles keyboard activation uniformly. Reuse the existing `ResponsiveMixin`'s `NARROW_BREAKPOINT` constant rather than hardcoding `768px` again here.

### 2. Mobile drawer accessibility is incomplete
**Severity:** HIGH • **Category:** accessibility/correctness • **Location:** `src/dorc-web/src/components/dorc-app.ts` (drawer styles + render method)

All three reviewers flagged distinct accessibility holes in the new mobile drawer:

- **No `inert` / `aria-hidden` when closed** (Opus): the closed drawer relies on `visibility: hidden; pointer-events: none; transform`, but the visibility transition is delayed 200 ms when closing, leaving keyboard users able to tab into hidden links during that window. There is also no focus trap, no Escape-to-close, and no `role="dialog"` / `aria-modal` when open.
- **Missing `aria-expanded` / `aria-controls` on the hamburger toggle** (Sonnet): `vaadin-button` carries only `aria-label="Toggle Menu"`, so screen readers cannot announce drawer state. WCAG 2.1 SC 4.1.2.
- **Visibility transition timing inconsistent + body scroll not locked when drawer is open** (Haiku): when opening, `visibility 0s` (no delay) means the drawer becomes focusable before the transform completes, and the page behind the drawer remains scrollable.

**Suggested fix:**
- Toggle an `inert` attribute (and `aria-hidden`) on `#dorcNavbar` based on open state on narrow screens.
- Bind `aria-expanded` (and `aria-controls="dorcNavbar"`) on the hamburger button to a reactive `_drawerOpen` state.
- When opening on narrow, also set `role="dialog" aria-modal="true" aria-label="Navigation"`, lock body scroll (`document.body.style.overflow = "hidden"`), trap focus, and add an Escape listener that closes the drawer.
- Make the visibility transition symmetric (delay both directions, or remove the delay and rely on `inert`).

---

**Other notable strong 2/3 findings worth a quick look:**

- **`:host` height conflict in many list pages**: the diff replaces the old `vaadin-grid#grid` selector with `:host`, leaving the original `height: calc(100vh - Npx)` in the same `:host` block alongside the new `height: 100%`. The later declaration wins, so the host is sized to the viewport instead of stretching to its flex parent — defeating the new flexbox contract and producing dual scrollbars on desktop, viewport overflow on mobile (100vh vs 100dvh). Affects `page-databases-list.ts`, `page-monitor-requests.ts`, `page-environments-list.ts`, `page-env-history.ts`, `page-config-values-list.ts`, `page-projects-list.ts`, `page-servers-list.ts`, etc. (Opus + Sonnet, **HIGH** correctness.)
- **Global `vaadin-combo-box` width override**: `registerStyles('vaadin-combo-box', :host { width: 100%; max-width: 600px; })` in `style-registrations.ts` applies to *every* combo box site-wide — including ones in toolbars, dialog headers, grid filter rows, and inline forms that previously sized to content. No opt-out hook. (Opus + Sonnet, **HIGH/MEDIUM** breaking.)
- **Splitter drag listeners leak**: `mousedown` adds `mousemove`/`mouseup` to `document.body`; if the component disconnects mid-drag, `mousemove` is never removed. (Sonnet + Haiku, **HIGH/MEDIUM** correctness.)
- **`page-about` uses `height: 100vh` on `#chart-all`** inside a flex parent that already constrains height — chart renders at full viewport height regardless of remaining flex space, producing a scrolling viewport-tall block. (Opus, **MEDIUM**.)
- **Test runner split** (`*.test.ts` for `@web/test-runner`, `*.vitest.ts` for Vitest) is undocumented and only one file (`page-monitor-requests.test.ts`) was renamed. A future contributor authoring a Vitest-style test under `.test.ts` will silently fail. CI's `npm test` may also only invoke one runner — verify both run. (Opus + Sonnet.)
- **Global 44×44 min size on icon buttons** (`registerStyles('vaadin-button', :host([theme~='icon']) { min-width: 44px; min-height: 44px; })`) applies on desktop too, enlarging dense toolbars and grid action buttons everywhere. Should be gated behind `@media (pointer: coarse)`. (Opus, **MEDIUM** breaking.)
- **New CSS variables `--dorc-header-height` and `--dorc-content-padding`** declared in `index.html` but never consumed — `dorc-app.ts` continues to use locally-scoped `--header-height: 50px`. The new canonical name is dead. (Sonnet, **MEDIUM**.)

---

.github/workflows/release.yml
```diff
@@ -68,7 +69,15 @@ jobs:
     - name: Build web site
       working-directory: src/dorc-web
       run: npm run build
-      
+
+    - name: Install Playwright browsers
+      working-directory: src/dorc-web
+      run: npx playwright install chromium
```
benhegartysefe: should we test for other browsers also? esp where the views are dynamic so we dont get different results

src/dorc-web/src/components/deploy/deploy-env.test.ts
```diff
@@ -0,0 +1,61 @@
+import { expect, fixture, html } from '@open-wc/testing';
```
benhegartysefe: All tests files hould be in a separate top level folder for the UI rather than next to the class file

src/dorc-web/vitest.config.ts
```diff
@@ -11,7 +11,7 @@ export default defineConfig({
   },
   test: {
     environment: 'jsdom',
-    include: ['src/**/*.test.ts'],
```
benhegartysefe, line 14: We seem to have 2 different web runner frameworks here, lets converge on 1

.github/workflows/release.yml
benhegartysefe: Done in 63d49c55 — CI now installs and runs against chromium, firefox and webkit.

src/dorc-web/src/components/deploy/deploy-env.test.ts
benhegartysefe: Done in 63d49c55 — all test files moved into top-level src/dorc-web/tests/ mirroring the src/ layout.

src/dorc-web/vitest.config.ts
benhegartysefe: Done in 63d49c55 — converged on vitest 4 with @vitest/browser-playwright; web-test-runner is removed.

src/dorc-web/src/components/dorc-app.ts
```diff
@@ -139,13 +196,54 @@ export class DorcApp extends ShortcutsStore {
...
+
   render() {
     return html`
       <header id="header" role="banner">
         <vaadin-button
           class="menu-btn"
           theme="icon"
           aria-label="Toggle Menu"
+          aria-controls="dorcNavbar"
+          aria-expanded="${this._drawerOpen ? 'true' : 'false'}"
```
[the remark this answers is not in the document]
hegsie: Addressed in c4bb366.

`aria-expanded` now reads from a new `_desktopSidebarVisible` state on the wide-screen branch (and from `_drawerOpen` on narrow), so the hamburger announces the actual visible state of the sidebar regardless of mode. Initial state is `true` on desktop (sidebar visible at `--dorc-sidebar-width` by default), and `toggleSideBar`'s desktop branch flips it alongside the inline `style.width`.

---

src/dorc-web/src/components/dorc-app.ts
```diff
@@ -201,35 +303,214 @@ export class DorcApp extends ShortcutsStore {
...
   }
 
+  private _openDrawer() {
+    if (!this.dorcNavbar) return;
+    this._drawerOpen = true;
+    this.dorcNavbar.classList.add('open');
+    if (this._narrowScreen) {
+      this._previouslyFocused = this._activeFocusedElement();
+      document.body.style.overflow = 'hidden';
+      this._drawerLockedScroll = true;
```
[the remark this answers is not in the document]
hegsie: Addressed in c4bb366.

`_openDrawer` now snapshots `document.body.style.overflow` into `_previousBodyOverflow` before setting it to `'hidden'`, and `_closeDrawer` restores that prior value (rather than unconditionally removing the property). Combined with the existing `_drawerLockedScroll` ownership flag from round-5, both the acquire and release sides of the scroll-lock now respect a coexisting overlay's state.

---

kirill-suspitsin-ext-sefe (approved):

PR #538, opened 2026-03-05
531/Use-1password-local-appsettings

### LINKED ISSUE #531: move deploysettings to 1password vault
we need to use 1password cli to retrieve values for app settings of local dev environments (monitor and API) from DORC the below vaults

- tr-devops-dorc-np
- tr-devops-dorc-pr

here is wiki with details on how to use 1password https://wiki/spaces/gdo/pages/745844974/How+To+-+Use+1Password+in+Development

slafeer-sefe:


benhegartysefe:

This PR was reviewed independently by 3 agents (Claude Opus 4.7, Sonnet 4.6, Haiku 4.5). No single finding cleared strict 3-way consensus (Sonnet alone went deep on the new `OnePasswordClient` code that the others didn't reach). Surfacing the **2-of-3** finding for awareness.

### `.gitignore` entries are no-ops because `appsettings.json` is already tracked
**Severity:** HIGH • **Category:** security • **Location:** `.gitignore` (~line 387–388)

Both `src/Dorc.Api/appsettings.json` and `src/Dorc.Monitor/appsettings.json` remain tracked in the PR branch (verifiable via `git ls-tree`). Git ignores `.gitignore` entries for files already in the index, so adding these paths to `.gitignore` does **not** prevent locally-edited (potentially secret-containing) `appsettings.json` from being staged or committed. The PR's stated security intent — preventing secrets from being committed — is not actually achieved. *Flagged by Opus + Sonnet.*

**Suggested fix:** In the same PR, run `git rm --cached src/Dorc.Api/appsettings.json src/Dorc.Monitor/appsettings.json` so the gitignore entries take effect, or commit a placeholder template file via `--skip-worktree`. Without removing the tracked files, the gitignore change is misleading.

---

**Other notable findings that did not reach 2-of-3 overlap but are worth a sanity-check:**

- `OnePasswordItem.GetFieldValue` returns the literal string `"Not Found"` instead of `string.Empty`/null when a field is missing. This propagates up through `GetSecretValueAsync` → `AddOAuth2Introspection`, so OAuth introspection silently uses `"Not Found"` as the client secret with no startup error. (Sonnet, **HIGH**.)
- No runtime mechanism resolves `op://` references in `appsettings.template.json`. The `op` CLI must be invoked manually before runtime; there's no MSBuild target, README update, or pre-build hook documenting the bootstrap. (Opus + Haiku, **MEDIUM**.)
- Hardcoded non-prod vault path (`tr-devops-dorc-np`) embedded across templates with no parameterisation; copy-paste mistakes preparing prod could mix non-prod secret refs with `IsProduction:false` defaults. (Opus, **MEDIUM**.)
- Sync-over-async (`.GetAwaiter().GetResult()`) on `GetSecretByItemId` blocks the calling thread during host startup; if the 1Password Connect service is slow or unreachable, startup stalls for up to the default 100s `HttpClient` timeout. (Sonnet, **MEDIUM**.)
- Deletion of `appsettings.Development.json` removes the only Development log-level override — the host won't auto-load `appsettings.template.json`. (Opus, **MEDIUM**.)

---
*Heuristic — these should be sanity-checked before action.*

Issue #231, opened 2025-07-16
Make Monitor service so it can be restarted without any issues with deployments


andrei-salanovich-ext-sefe:
Today when deployment requests are running, and Monitor service is being restarted, some requests are hang forever in 'Running' state which cause all others requests for the same environment wait forever.

Design and implement a mechanism that allows the service to restart seamlessly without affecting deployment requests.

dmitryrext:
this has been fixed in other releases

PR #544, opened 2026-03-09
feat: graceful shutdown and orphan recovery for Monitor service (#231)

smurugan-sefe:


benhegartysefe:

### 1. Fire-and-forget event publishing in `RecoverRequestsForEnvironment` swallows exceptions
**Severity:** HIGH • **Category:** correctness • **Location:** `src/Dorc.Monitor/DeploymentRequestStateProcessor.cs` (`RecoverRequestsForEnvironment`, ~line 180–188)

The recovery loop publishes `Pending` status-changed events using `_ = this.eventPublisher.PublishRequestStatusChangedAsync(...)` — discarding the returned `Task` with no `try/catch` and no awaiter. Any exception from the publisher is unobserved (subject to `TaskScheduler.UnobservedTaskException` policies, which can crash the process), and any failure to publish is silently lost.

The same file already provides `PublishRequestStatusChangedSafe()` (recently merged on `main`) that wraps the call in `Task.Run` + `try/catch` + logging, precisely to avoid this. Reusing the unsafe pattern in new code is inconsistent with the surrounding file and will create merge conflicts when this PR is rebased onto current `main`.

**Suggested fix:** Use `PublishRequestStatusChangedSafe()` (or equivalent: `await` the call inside an async loop and add a per-iteration try/catch that logs failures). Apply the same gating to any other fire-and-forget publish sites introduced by the PR.

---

PR #558, opened 2026-03-17
Fix PS7 Progress stream InvalidCastException by adding dedicated Powe…

### LINKED ISSUE #552: Some scripts fail in PS7
When running with PS7 in QA env controllservices.ps1 fails with following error (deployment ID - 1816135):
"[ERR] [PS] ErrorId: Microsoft.PowerShell.Commands.WriteErrorException,Get-DorcToken
Exception: Microsoft.PowerShell.Commands.WriteErrorException: Failed to retrieve token. Please check DorcIDSsecret variable value! Error: Unable to cast object of type 'System.Management.Automation.PSDataCollection`1[System.Management.Automation.ProgressRecord]' to type 'System.Management.Automation.PSDataCollection`1[System.Management.Automation.InformationRecord]'."

Based on my investigation the cause of the issue is following :
Wrong event wiring in DOrc.Cmd/dorc/src/Dorc.PowerShell/PowerShellScriptRunner.cs: Progress stream is subscribed to the Information handler.
Specifically line 61: powerShell.Streams.Progress.DataAdded += Powershell_Information_DataAdded;

smurugan-sefe:
…rshell_Progress_DataAdded handler

### DIFF
--- a/src/Dorc.PowerShell.Tests/Dorc.PowerShell.Tests.csproj
+++ b/src/Dorc.PowerShell.Tests/Dorc.PowerShell.Tests.csproj
@@ -1,0 +1,24 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <LangVersion>latest</LangVersion>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="18.0.0" />
+    <PackageReference Include="MSTest" Version="4.0.1" />
+    <PackageReference Include="NSubstitute" Version="5.3.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Using Include="Microsoft.VisualStudio.TestTools.UnitTesting" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Dorc.PowerShell\Dorc.PowerShell.csproj" />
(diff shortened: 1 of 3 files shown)


benhegartysefe:

### 1. Test assertion can pass vacuously — does not actually verify the fix
**Severity:** MEDIUM • **Category:** test • **Location:** `src/Dorc.PowerShell.Tests/PowerShellScriptRunnerTests.cs`

All three reviewers flagged that the assertion `mockLogger.Received().Information(Arg.Is<string>(s => s.Contains("[PS]")))` can be satisfied by the `Write-Output 'Done'` call alone — it does **not** prove that progress records were processed without throwing, which is the regression the PR claims to fix. NSubstitute's `Received()` (no count argument) only checks for at-least-one matching call across the entire test, so an `InvalidCastException` raised on the event-handler thread (caught by the outer try/catch and returned as `-1`) would still let the assertion pass.

Additionally, both reviewers flagged that these are **integration tests masquerading as unit tests**: they spin up a real PowerShell runspace, write temp `.ps1` files, and depend on PowerShell 7 being installed on the test host (per project memory, ARM macOS doesn't have it). They will fail or be silently skipped in unsupported environments, with no `[TestCategory("Integration")]` to gate them.

**Suggested fix:**
- Assert on the *progress*-specific text (e.g. `Arg.Is<string>(s => s.Contains("Step 1") || s.Contains("Testing"))`) so the assertion fires only when the new handler ran. Optionally also assert `mockLogger.DidNotReceive().Error(...)` to catch silent failures.
- Mark the tests `[TestCategory("Integration")]` (or move them to a dedicated integration project, e.g. `Dorc.PowerShell.IntegrationTests`) and document the PS7 dependency so CI can gate appropriately.

---

Issue #563, opened 2026-03-23
when you clone environment with properties, they should be altered as well


dmitryrext:
some environment's properties **values** contain environment name in them
those should be altered when you clone the environment, such properties values should contain new env name

PR #564, opened 2026-03-23
Replace source env name in property values during clone (#563)

smurugan-sefe:


benhegartysefe:

### 1. Substring `Replace` can corrupt values where the env name is part of a larger token
**Severity:** HIGH • **Category:** correctness • **Location:** `src/Dorc.PersistentData/Sources/EnvironmentsPersistentSource.cs` (~line 1198–1201)

`string.Replace(sourceEnvName, destEnvName, OrdinalIgnoreCase)` performs a substring match anywhere in the value, with no word-boundary check. For short or common env names (`DEV`, `UAT`, `QA`) this fires on tokens like `DEVELOPER_KEY`, `\\server\DEV\share`, `dev-router-prod.corp`, etc. Cloning `DEV → PROD` would silently rewrite `DEVELOPER_KEY` to `PRODELOPER_KEY` and produce broken configuration in the cloned environment. Case-insensitive matching also means embedded mixed-case occurrences lose their original casing in URLs, file paths, and other case-sensitive identifiers.

**Suggested fix:** Match the env name only when it appears as a delimited token — either via `Regex.Replace` with `\b…\b`, or by checking that surrounding characters are non-alphanumeric. Document the chosen heuristic alongside the change.

### 2. New replacement logic has no test coverage
**Severity:** MEDIUM • **Category:** test • **Location:** `src/Dorc.Api.Tests/EnvironmentsPersistentSourceTests.cs` (no tests for `CopyPropertyValues`)

There is no test for `CopyPropertyValues` at all, so the new substring-rewrite branch (and its OrdinalIgnoreCase semantics) ships with zero regression coverage. Combined with the substring-collision risk above, future env-naming changes can silently regress.

**Suggested fix:** Add unit tests covering at least: (1) exact-match replacement, (2) no-match passthrough, (3) null/empty value passthrough, (4) value where env name is a substring of a larger token, and (5) source name == dest name.

---

PR #597, opened 2026-04-09
112-api access audit logs

slafeer-sefe:


benhegartysefe:

### 1. Hardcoded Windows-only audit log path
**Severity:** HIGH • **Category:** architecture • **Location:** `src/Dorc.Api/Security/ApiAccessAuditMiddleware.cs` (~line 14–20)

The audit log path is hardcoded to `c:\Log\DOrc\Deploy\Web\Api\audit\Dorc.Api.Audit.log`. This breaks any non-Windows deployment (notably the Azure Container Apps demo, which runs on Linux), prevents per-environment configuration, and will throw at first request if the directory cannot be created. The existing main `Dorc.Api.log` sink is driven by `loggerSettings.json` — the new audit sink ignores that pattern entirely and is also invisible to ops staff who manage logging via configuration.

**Suggested fix:** Move the path and rolling interval into `loggerSettings.json` (as a separate Serilog logger or sub-logger keyed by `SourceContext`/filter), or read it from `IConfiguration` with a sensible cross-platform default. Inject the configured logger into the middleware rather than constructing one inline.

### 2. Serilog logger constructed in middleware constructor — file handles leak, host shutdown loses entries
**Severity:** HIGH • **Category:** correctness • **Location:** `src/Dorc.Api/Security/ApiAccessAuditMiddleware.cs` (~line 10–21)

`new LoggerConfiguration()…CreateLogger()` is invoked in the constructor and the resulting `Logger` (which owns a file sink and background workers) is never `Dispose()`d. The dedicated logger sits outside the application's `Log.CloseAndFlush()` / host shutdown flow, so the final batch of audit entries can be lost on graceful shutdown. If anything ever changes the middleware lifetime (transient registration, test pipeline rebuilds), it also leaks file handles.

**Suggested fix:** Either inject `ILogger<ApiAccessAuditMiddleware>` (resolved via DI, like `WinAuthLoggingMiddleware` next to it does), or wire the audit sink into the central Serilog pipeline so it benefits from the host's `CloseAndFlush` lifecycle.

### 3. Audit log entry skipped when downstream middleware throws
**Severity:** HIGH • **Category:** correctness • **Location:** `src/Dorc.Api/Security/ApiAccessAuditMiddleware.cs` (~line 22–39)

`await _next(context)` is not wrapped in `try/finally`. Any unhandled exception from a downstream component (controller, auth, exception handler) bypasses the audit write entirely, so requests that error — exactly the ones an access audit needs to capture — are silently omitted from the log.

**Suggested fix:** Wrap `_next(context)` in `try/finally` and emit the audit record from the `finally` block. Optionally annotate the entry with whether the response completed normally vs. faulted.

### 4. Middleware bypasses DI and the central Serilog/OTel pipeline
**Severity:** MEDIUM • **Category:** architecture • **Location:** `src/Dorc.Api/Security/ApiAccessAuditMiddleware.cs`

`Program.cs` configures a single Serilog logger plus an OpenTelemetry OTLP exporter for centralised log shipping. The new middleware constructs its own isolated Serilog logger that writes only to a local file, so audit events are not exported via OTLP, not enriched with shared context (`ThreadId`, etc.), and not subject to the configured `MinimumLevel` overrides. Operators expecting all API logs to flow through the central pipeline will silently miss audit records. The middleware also can't be unit tested because it owns its dependency.

**Suggested fix:** Use a `SourceContext`-tagged `ILogger` from the central Serilog instance (e.g. `Log.ForContext<ApiAccessAuditMiddleware>()`) and route audit records to a sub-logger via Serilog filtering in `loggerSettings.json`. This preserves OTLP export and shared enrichers — and lets the middleware be mocked in tests.

---

PR #599, opened 2026-04-10
Feature/595 frontend tests

### LINKED ISSUE #595: Add frontend unit tests for dorc-web components
## Background
The dorc-web project has no unit test infrastructure. All UI behavior is manually tested. PR #584 added significant UI logic (source control type icons, combo-box binding, conditional field visibility) that would benefit from automated tests.

## Current State
- No test framework configured (`"test": "echo no test specified"`)
- Only k6 load tests exist (`k6-tests/`) which test API performance, not UI logic
- `page-test-component.ts` exists but is a development helper, not a test

## Recommended Approach
1. Add Vitest (works with Vite out of the box) + `@open-wc/testing` for Lit component testing
2. Priority components to test:
   - **`add-edit-project.ts`** — source control type combo-box binding, conditional field visibility (showSubPaths/showBuildRegex), label switching per type
   - **`page-projects-list.ts`** — source control icon renderer (AzDO/GitHub/FileShare detection), project name tooltip, search filtering
   - **`project-controls.ts`** — button visibility based on permissions
   - **`deploy-env.ts`** — build definition handling, folder project detection
3. Test the enum string serialization edge case (API returns `"FileShare"` not `2`)

## Setup Steps
```bash
npm install -D vitest @open-wc/testing @testing-library/dom jsdom
```

Add to `vite.config.ts`:
```ts
test: {
  environment: 'jsdom',
  include: ['src/**/*.test.ts']
}
```

_Identified during PR #584 UI development_

smurugan-sefe:


benhegartysefe:

This PR was reviewed independently by 3 agents (Claude Opus 4.7, Sonnet 4.6, Haiku 4.5). Only findings where **all three** independently flagged the same code area are posted.

### 1. `StubXMLHttpRequest` in `test-setup.ts` is defective
**Severity:** MEDIUM • **Category:** test-quality • **Location:** `src/dorc-web/src/test-setup.ts` (~line 17–29)

All three reviewers independently flagged the new XHR stub as broken, each from a different angle:

- **Missing `super()`:** the stub `extends XMLHttpRequest` without calling `super()`. Browser APIs with non-configurable internal slots can't be subclassed this way; it works in `happy-dom` only because that environment is permissive. (Haiku, **HIGH**.)
- **Scope is too broad:** the comment says it targets `appconfig.json` loading, but the implementation intercepts *every* XHR for *every* test. Any future test that relies on XHR or any production code path that surfaces XHR errors will silently succeed, masking real failures. (Opus, **MEDIUM**.)
- **Never invokes callbacks:** `send()` sets `status = 0` as an own property but never calls `onerror`, `onloadend`, or `onreadystatechange`. Any constructor that awaits an XHR response will silently hang or initialise to a half-state in tests, with no failure signal. (Sonnet, **MEDIUM**.)

The combined effect is a stub that gives confidence in test passes which may not reflect real behaviour.

**Suggested fix:** Either replace the stub with a scoped `vi.spyOn` on `window.fetch` / `XMLHttpRequest.prototype.open` that only intercepts the specific `appconfig.json` URL and dispatches a synthetic event so error-handling paths run; or, preferably, refactor `app-config.ts` to take a fetcher as a dependency so tests can inject a mock without touching globals.

---

**Other notable findings that fell short of strict 3-way consensus** (worth considering but flagged by only 2/3 reviewers):

- The `unhandledRejection` listener replacement (lines 31–46) is installed at module load with no teardown and uses broad keyword matching (`'lumo'`, `'ECONNREFUSED'`); a real production rejection containing those substrings would be silenced. (Opus + Sonnet.)
- `convertDeployCompToTree` test asserts `hasParent: true` for `{ParentId: 5, NumOfChildren: 3}` — the source computes `hasParent` from `numChildren`, not `parentId`. The test certifies the bug. A discriminating case (`ParentId > 0, NumOfChildren = 0`) would expose it. (Opus + Sonnet, **HIGH**.)
- `jsdom` (~8 MB) added to devDependencies but never imported — vitest is configured to use `happy-dom`. (Opus + Sonnet.)
- `globalThis.alert` is replaced with `vi.fn()` inside `it` blocks without restoration — leaks across subsequent tests. (Opus + Sonnet.)

---

PR #610, opened 2026-04-14
remove env owner info from metadata page

slafeer-sefe:
all related code removed

### DIFF
--- a/src/dorc-web/src/components/add-edit-environment.ts
+++ b/src/dorc-web/src/components/add-edit-environment.ts
@@ -1,25 +1,18 @@
-import { css, LitElement, PropertyValues, render } from 'lit';
+import { css, LitElement, PropertyValues } from 'lit';
 import '@vaadin/grid/vaadin-grid-sort-column';
 import '@vaadin/grid/vaadin-grid';
 import '@vaadin/text-field';
-import '@vaadin/combo-box';
 import '@vaadin/button';
 import '@vaadin/details';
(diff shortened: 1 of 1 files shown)


benhegartysefe:

### 1. New environments will persist `EnvironmentOwner = 'NotSet'` to the DB
**Severity:** HIGH • **Category:** correctness • **Location:** `src/dorc-web/src/components/add-edit-environment.ts` (`getEmptyEnv`, ~line 326–343)

`getEmptyEnv()` still seeds `EnvironmentOwner = 'NotSet'` and `EnvironmentOwnerId = ''`, but the validation in `_inputValueChanged()` that previously blocked submission while the owner was `'NotSet'`/empty has been removed alongside the UI. As a result, every environment created through the Add flow now ships with the literal sentinel `'NotSet'` as its owner, which propagates into `EnvironmentsPersistentSource.MapToEnvironment` and creates a bogus Owner-level AccessControl row (`Name='NotSet'`, `Pid=''`, `Sid=null`). This breaks owner-based authorization and any owner-email enrichment downstream.

**Suggested fix:** Either auto-assign the creating user as the initial owner on the server when `EnvironmentOwnerId` is empty, or skip adding the Owner AccessControl row when the owner is null/empty/`'NotSet'`. At minimum, change `getEmptyEnv()` to default `EnvironmentOwner` to `null`/`''` and reject `'NotSet'` server-side.

---

PR #615, opened 2026-04-16
add discover daemons endpoint and admin only frontend button

### LINKED ISSUE #560: some of daemons not showing in DORC
how to reproduce: 
1. Open environment, mentioned in PBI 412430
2. Go to "daemon" tab and click "load"
3. Only one daemon will be visible, but there should be more

slafeer-sefe:


benhegartysefe:

### 1. New `/discover` endpoints have no server-side admin authorization
**Severity:** HIGH • **Category:** security • **Location:** `src/Dorc.Api/Controllers/DaemonStatusController.cs` (`DiscoverForEnvironment` and `DiscoverForAllEnvironments`, ~line 104–157)

Both `POST /DaemonStatus/discover` and `POST /DaemonStatus/discover/{envName}` are gated only by `[Authorize]`. The PR title and frontend describe the action as admin-only — the UI hides the button via `?hidden="${!this.isAdmin}"` — but the API itself never calls `IRolePrivilegesChecker.IsAdmin(User)`. Any authenticated user can invoke the endpoints directly (curl/Postman), triggering an impersonated `WindowsIdentity` logon, network probes against every server in the environment(s), and persisted writes to daemon-to-server mappings (`PersistDiscoveredMappings`). Other admin-only controllers in this codebase explicitly check `IsAdmin` — e.g. `RefDataConfigController`.

**Suggested fix:** Inject `IRolePrivilegesChecker` and add `if (!_rolePrivilegesChecker.IsAdmin(User)) return Forbid();` at the top of both action methods (or `[Authorize(Roles = "Admin")]`). For the per-env variant, also call `_securityPrivilegesChecker.CanModifyEnvironment` for parity with `PutDaemonState`.

---

PR #619, opened 2026-04-20
99 - fix high sonarqube issues

### LINKED ISSUE #99: 338812 - Fix the SonarQube Issues in Dorc
<div><a href="https://sonarqube/project/issues?resolved=false&amp;impactSeverities=HIGH&amp;types=BUG&amp;id=sh.devops.dorc">https://sonarqube/project/issues?resolved=false&amp;impactSeverities=HIGH&amp;types=BUG&amp;id=sh.devops.dorc</a><br> </div><div><br> </div><div><br> </div><div><br> </div>

slafeer-sefe:


benhegartysefe:

This PR was reviewed independently by 3 agents (Claude Opus 4.7, Sonnet 4.6, Haiku 4.5). Haiku found no concerns; Opus and Sonnet flagged distinct concrete regressions. No single issue cleared strict 3-way consensus, but the **2-of-3** finding below is worth surfacing because it is a quiet but high-impact regression for a sweep PR.

### `GetPermissions(int)` parameter renamed `databaseId` → `userId` — but the body still filters on `DbId`
**Severity:** MEDIUM • **Category:** correctness • **Location:** `src/Dorc.PersistentData/Sources/UserPermsPersistentSource.cs` (~line 19–29 / 6973)

The single-arg overload was renamed, but the EF query is still `environmentUser.DbId == userId`. Callers split:

- `VariableScopeOptionsResolver.cs:157` passes `databaseApiModel.Id` — matches the *real* semantics, contradicts the new name.
- `ManageUsers.cs:38` passes a `userId` — matches the new name, contradicts the semantics. This is likely a pre-existing bug now hidden by the rename.

The runtime behaviour is unchanged (the WHERE clause was not touched), but the rename codifies the wrong contract: any future caller reading the signature will believe they should pass a user ID and silently get the wrong rows. *Flagged by Opus + Sonnet.*

**Suggested fix:** Revert the parameter name to `databaseId` (truthful to the WHERE clause), and audit `ManageUsers.cs:38` separately to confirm/fix what it actually intends to query.

---

**Other notable single-reviewer findings worth a quick sanity-check** (Opus and Sonnet flagged different specific regressions across the sweep — sharing here in case any resonate):

- **From Opus** — `DeploySingleComponent` returns `null` for `OperationCanceledException` and discards the computed `Cancelled`/`Pending` status; user-cancelled requests are persisted as Completed. Also: `DeployComponents` always overwrites `deploymentRequestStatus` per iteration, so a successful later component overwrites a prior `Failed` (when `StopOnFailure=false`) or `WaitingConfirmation` set by a Terraform component. (Both **HIGH** correctness.)
- **From Sonnet** — `AsyncStreamReader.EnqueueLine` refactor moves the `isMessageLast` field assignment outside the `lock(messageQueue)` block, opening a race window between sentinel detection and consumer thread reading the field. Also: `Marshal.GetLastWin32Error()` is called inside the `Win32Exception` constructor in `TerraformDispatcher`, which invalidates the subsequent `Marshal.GetLastWin32Error()` check in the same scope. (Both **HIGH** correctness.)

These deserve a closer look since they were flagged with `HIGH` severity by the discovering model — even without 2/3 overlap, a sweep PR is exactly the place where a single subtle regression can survive into main.

---
*Heuristic — this is a hint, not a verdict; sanity-check before acting.*

PR #642, opened 2026-04-22
Add dark mode to Swagger UI #641

### LINKED ISSUE #641: Swagger dark mode
Switching between DOrc's dark mode UI and the default bright Swagger page creates a harsh contrast. Added dark mode styling to Swagger UI for a consistent experience. Includes auto-detection of system theme preference and a manual toggle.

smurugan-sefe:


benhegartysefe:

### 1. `UseStaticFiles()` placement is problematic
**Severity:** MEDIUM • **Category:** security/architecture • **Location:** `src/Dorc.Api/Program.cs` (~line 340)

All three reviewers flagged the new `app.UseStaticFiles()` call, framing the same root cause from different angles:

- It serves the entire `wwwroot/` folder to anonymous callers — today only the two Swagger files exist there, but any future file accidentally dropped in (config dumps, generated reports, certs) becomes publicly downloadable without a code change.
- It is registered **before** `UseAuthentication()` / `UseAuthorization()`, so static-file responses bypass any auth scheme configured for the rest of the API.
- Its position relative to other middleware (e.g. `UseExceptionHandler`) is non-standard for ASP.NET Core pipelines.

Given the API hosts deployment-orchestration endpoints, the blast radius of a future mistake here is wider than a two-file Swagger-theming feature warrants.

**Suggested fix:** Scope the static-file middleware to only the two assets — e.g. `UseStaticFiles(new StaticFileOptions { RequestPath = "/swagger-assets", FileProvider = new PhysicalFileProvider(...) })` mapped to a dedicated subfolder. Alternatively, use `SwaggerUIOptions.InjectStylesheet` / `InjectJavascript` overloads that accept content directly (no static-file middleware needed).

---

kirill-suspitsin-ext-sefe (approved):

PR #659, opened 2026-04-29
#448 Add terraform runner logging in reaaltime

### LINKED ISSUE #448: Terraform component runs full deployment(all components) after applying
**Steps**
- start deployment using Terraform component and any other (include several components to deployment)
- after Terraform plan will be created, approve it and proceed

**Actual result**
The component, different from the one was approved (Terraform one), starts to run

**Expected result**
Only approved Terraform component starts running, not affecting others

kirill-suspitsin-ext-sefe:
Add terraform runner logging in reaaltime

### DIFF
--- a/src/Dorc.Api/Controllers/TerraformController.cs
+++ b/src/Dorc.Api/Controllers/TerraformController.cs
@@ -36,6 +36,7 @@ namespace Dorc.Api.Controllers
             _azureStorageAccountWorker = azureStorageAccountWorker;
         }
 
+
         /// <summary>
         /// Gets Terraform plan for a deployment result
         /// </summary>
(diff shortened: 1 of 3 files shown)


benhegartysefe:

### 1. Stderr is still buffered — defeats the "real-time" goal
**Severity:** MEDIUM • **Category:** correctness • **Location:** `src/Dorc.TerraformRunner/TerraformProcessor.cs` (`ErrorDataReceived` handler, ~line 202–208)

Only the `OutputDataReceived` (stdout) handler now calls `logger.Information(...)` per line. The `ErrorDataReceived` (stderr) handler still only appends to `errorBuilder` and is never emitted in real-time — it's only logged at all when `ExitCode == 1`. Terraform writes warnings, deprecation notices, provider-install progress, and many diagnostic messages to stderr, so the "real-time" stream the PR promises is missing exactly the lines that operators most need when a `plan`/`apply` is misbehaving. With the consolidated post-completion `Output:` log line also removed (line ~261), stderr produced during a successful run (Terraform frequently emits warnings while exiting 0/2) is now dropped from both file logs and OpenSearch.

**Suggested fix:** Mirror the new `logger.Information($"terraform:{e.Data}")` call inside `ErrorDataReceived` (e.g. `logger.Warning($"terraform-error:{e.Data}")`) so stderr is also surfaced as it is produced.

---

PR #660, opened 2026-04-29
655 - implement audit for servers and databases similar to daemons

### LINKED ISSUE #655: Database and Server audit
As we already have the Audit page for projects, daemons, properties and history page for the environments.
We also need to add the new Audit functionality for the Databases and Servers, that should include creation, metadata editing, linking to environments and linking with daemons for servers.
UI should be the same as for Daemons (check main branch locally, this functionality hasn't been deployed on prod yet)

slafeer-sefe:


benhegartysefe:

This PR was reviewed independently by 3 agents (Claude Opus 4.7, Sonnet 4.6, Haiku 4.5). No single finding cleared strict 3-way consensus — each model focused on a different layer (Opus on infrastructure/perf, Sonnet on the controller mutation logic, Haiku on schema FKs). Surfacing the strongest **2-of-3** findings for awareness.

### 1. `Put` audit captures the wrong "before" state — `fromValue` is null on every successful update
**Severity:** HIGH • **Category:** correctness • **Location:** `src/Dorc.Api/Controllers/RefDataDatabasesController.cs` (~line 240–247)

In `Put`, `databaseApiModel` is the result of `GetDatabase(database.Id)` fetched earlier specifically to detect a duplicate-name *collision*: it returns the *other* record that already has the requested name, not the record being updated. When there's no collision (the normal path), this variable is `null`, so `beforeJson` is always `null` and the audit row records no before-state for every successful update. The server controller has a structurally similar issue: it calls `GetServer(server.Name, User)` (by name, for the collision check) and reuses that as the before-state snapshot. *Flagged by Sonnet + Haiku.*

**Suggested fix:** Fetch the before-state explicitly by ID before performing the update — `var beforeDatabase = _databasesPersistentSource.GetDatabase(id);` — and serialise that. Do not reuse the collision-check variable.

### 2. `WriteComponentAudit` silently does nothing for unrecognised component types
**Severity:** MEDIUM • **Category:** correctness • **Location:** `src/Dorc.Api/Controllers/RefDataEnvironmentsDetailsController.cs` (~line 316–323)

The switch on `component.ToLowerInvariant()` only handles `"server"` and `"database"`. Any other value (typo, future extension) falls through with no audit and no error. Since the call site maps raw API input to this switch, gaps will be introduced silently. *Flagged by Sonnet + Haiku.*

**Suggested fix:** Add a `default` case that throws `ArgumentOutOfRangeException` (or at minimum logs a warning) so future callers are not silently unaudited.

### 3. Audit writes are not transactional with the mutating operation or with each other
**Severity:** MEDIUM • **Category:** correctness • **Location:** `ServerDaemonsController.cs` (Attach/Detach, ~line 83–141), and the controller mutation paths generally

Attach/Detach now performs three independent writes against three sessions: the mapping change, then `InsertDaemonAudit` (own `DbContext` + `SaveChanges`), then `InsertServerAudit` (own `DbContext` + `SaveChanges`). Any failure between them leaves audit partially written: the mapping change can be persisted with no audit, or only one of the two audit rows present. In the Delete path the risk is reversed — `DeleteDatabase` succeeds, then the audit insert throws and the user gets a 500 even though the delete already happened. *Flagged by Opus + Sonnet.*

**Suggested fix:** Either share a single `DbContext`/transaction across the mutation + both audit inserts, or wrap audit writes in `try/catch + log` so audit failures don't override a successful user-facing action.

---

**Other notable single-reviewer findings worth a quick sanity-check:**

- **Audit endpoints expose full history to any authenticated user** (Opus, **MEDIUM** security): `ServerAuditController` and `DatabaseAuditController` carry only `[Authorize]`, no role check. Any authenticated user can read every server/database mutation, including before/after JSON snapshots and admin logon names. The mutating endpoints these audits cover require write/PowerUser; the audit *reader* does not.
- **Missing FK constraints to `Servers`/`Databases`** (Haiku, **HIGH**): `ServerAudit.ServerId` and `DatabaseAudit.DatabaseId` have no FK, allowing orphaned audit rows.
- **`PUT` used for a read/query operation** in both new audit controllers — wrong HTTP semantics, breaks REST. (Sonnet, **HIGH** architecture.)
- **`DateTime.Now` for audit timestamps** — local-time ambiguity across hosts/DST. (Sonnet, **MEDIUM**.)
- **Audit tables lack indexes for the default `OrderByDescending(Date)` query path** — full table scans as audit grows. (Opus, **MEDIUM**.)
- **`WhereAll` returns empty set when predicate list is empty** (`Where(x => false)` instead of returning the source). Currently unreachable but a footgun for future callers. (Sonnet, **MEDIUM**.)

---
*Heuristic — even with 2/3 overlap, these should be sanity-checked before action.*

kirill-suspitsin-ext-sefe (approved):

PR #669, opened 2026-05-06
feature - MS Teams bot DM notifications on completion of deployment

### LINKED ISSUE #657: MS Teams bot DM notifications on deployment completion (Bot Framework proactive messaging)
## Problem

When a deployment request finishes, the only notification surface is the SignalR-driven real-time UI. A user who isn't actively watching the page has no way of knowing their job has finished. We want to push a **direct 1:1 MS Teams message** to the requesting user when their request reaches a terminal state.

## Goals

- Deliver a Teams **direct message (1:1 chat)** to the user who submitted a deployment request when that request reaches a terminal state.
- Message is an **Adaptive Card** containing: request id, requester, project, environment, build number, final status, start/completed timestamps, duration, and a deep-link back to the DOrc UI (`/monitor-result/{id}`).
- Failure-tolerant: notification errors must NOT affect deployment status persistence or other event publishing.
- Configurable on/off per environment via `appsettings.json` (`TeamsNotification` section) and the MSI installer parameters.

## Non-goals (this issue)

- Two-way Teams interaction (approving WaitingConfirmation requests from Teams). Track separately.
- Email / Slack / generic webhook channels. The `IDeploymentNotificationSink` abstraction leaves room for them later.
- Per-user opt-in/opt-out preferences UI. v1 sends to the requester unconditionally.
- Channel / Team broadcast notifications. v1 is strictly user-targeted DMs.

## Delivery model: Bot Framework proactive messaging (no Graph)

> ⚠️ This section supersedes the original Graph-based design. The Graph chat approach (`Chat.Create` / `ChatMessage.Send` + MSAL) ran into problems in our tenant and was abandoned — see [comment](https://github.com/sefe/dorc/issues/657#issuecomment-4507601172). No Graph chat permissions are required.

The Monitor sends proactive messages directly through the Bot Framework Connector:

```
PendingRequestProcessor / DeploymentRequestStateProcessor reach terminal status
  → IDeploymentNotificationSink.NotifyRequestCompletedAsync(request, finalStatus, startedTime, completedTime)
    → TeamsBotNotificationSink: NotifyOnStatuses filter → resolve request.UserName → Entra object id (AzureEntraSearcher)
      → ITeamsConversationClient.CreateConversationAsync (tenant + AAD object id)
      → ITeamsConversationClient.SendCardAsync (Adaptive Card 1.4)
```

**What this design requires:**

1. **Azure AD app registration + Azure Bot Service** resource with the Microsoft Teams channel enabled — ✅ done (registrations created; currently configured for ut2 only).
2. **The bot app installed for the user in Teams** — this is the only per-user prerequisite for receiving proactive DMs.
3. Bot credentials (`BotAppId` / `BotAppPassword` / `TenantId`) and the regional Connector `ServiceUrl` (`https://smba.trafficmanager.net/uk/`) in configuration; secrets sourced from the secret store, never committed.

**What this design explicitly does NOT need** (all removed from the original plan):

- ❌ Microsoft Graph chat permissions (`Chat.Create`, `ChatMessage.Send`, `User.Read.All` for chat) / MSAL client-credentials flow
- ❌ `TeamsUserConversation` conversation-reference table in `Dorc.Database`
- ❌ Bot Framework activity endpoint (`TeamsBotController`) in `Dorc.Api`
- ❌ A separate `Dorc.Notifications.Teams` project — components live under `src/Dorc.Monitor/Notifications/`

## User identity resolution

- `DeploymentRequestApiModel.UserName` is populated from the requester's claims (`OAuthClaimsPrincipalReader`): usually the **email claim**, falling back to SAM account name; for M2M requests it can be a machine client id.
- `TeamsBotNotificationSink` resolves `UserName` → Entra **object id** via the existing `AzureEntraSearcher.Search()` in `Dorc.Core` (Graph *user lookup* under the existing `AppSettings.Aad*` app credentials — distinct from the rejected Graph *chat* APIs).
- Resolution failure (or empty `UserName`) → log a warning and skip the notification; the deployment is never affected.

## Implementation status — PR #669 (updated 2026-08-02)

v1 is implemented in #669 (branch `feature/teams-notification-bot`), including the gap-closure work below:

- `src/Dorc.Monitor/Notifications/` — `IDeploymentNotificationSink`, `NoOpDeploymentNotificationSink`, `DeploymentNotificationDispatch` (fire-and-forget swallow policy), `FatalExceptions`
- `src/Dorc.Monitor/Notifications/Teams/` — `TeamsBotNotificationSink` (status filter, resolution, Polly retry + cooperative 10s timeout; 4xx/timeout/cancellation not retried), `ITeamsConversationClient`/`TeamsConversationClient` (Bot Connector wrapper), `DeploymentCompletionCardBuilder`, `TeamsBotOptions`
- Hookpoints: `PendingRequestProcessor` terminal sites **and** `DeploymentRequestStateProcessor` (cancel / abandon / stale-cleanup / errored-pickup), per-request optimistic transitions so HA monitor pairs never double-DM
- DI + options in `src/Dorc.Monitor/Program.cs` (bound via `builder.Configuration` so env-var/secret overrides work; malformed `Enabled` disables rather than crashing)
- Installer wiring: `Setup.Dorc` (bat, NonProd/Prod wxs, msi.json) for all 7 `TEAMS.*` parameters
- Tests: 30+ unit tests (card builder, sink, both processors' notification behaviour) in `src/Dorc.Monitor.Tests`
- `teams-app/` manifest template + placeholder icons; `docs/teams-notifications/teams-notifications-setup.md` setup & rollout guide

Configuration (`src/Dorc.Monitor/appsettings.json`):

```json
"TeamsNotification": {
  "Enabled": "false",
  "BotAppId": "",
  "BotAppPassword": "",
  "TenantId": "",
  "ServiceUrl": "https://smba.trafficmanager.net/uk/",
  "DorcUiBaseUrl": "http://localhost:8888",
  "NotifyOnStatuses": "Completed,Failed,Errored"
}
```

## Remaining work for v1 (tracked on PR #669)

- [x] **Monitor-side Entra config + DI** — the `AppSettings.Aad*` keys were already present in the Monitor's appsettings and installer; the searcher is now injected via DI instead of constructed inside the sink.
- [x] **Tests** — unit coverage for card builder, sink (disabled / filter / empty user / resolution failure / retry / swallowed failures) and both processors (fires exactly once per terminal status; faulted sink tasks never affect completion).
- [x] **Code-quality findings from PR #669** — obsolete `TrustServiceUrl` removed, conversation id null-guarded, generic catches narrowed to non-fatal filters; stray `src/src/...` duplicate deleted.
- [x] **`NotifyOnStatuses` filter** — configurable, default `Completed,Failed,Errored`; effective set logged at startup, unknown names warned.
- [x] **Cancel/abandon path coverage** — `DeploymentRequestStateProcessor` now notifies on Cancelled/Abandoned/Errored transitions (per-request, duplicate-safe in HA; `PendingRequestProcessor` defers Cancelled to it).
- [ ] **Rollout** — manifest template + setup docs are in the repo; still to do operationally: real icons + privacy/terms URLs in the manifest, tenant catalog upload (U3), enable environments beyond ut2.

## Deferred (follow-up candidates)

- Mark `TEAMS.BOT.APP.PASSWORD` `Hidden="yes"` in the WiX property definitions so it stays out of verbose MSI logs.
- Decide notification behaviour for M2M/machine requesters (currently fails safe: skip + log).
- Card "Duration" uses queue-wait for never-started requests (cancel/abandon paths) — cosmetic.

## Acceptance criteria

- [ ] When a deployment request reaches a status listed in `NotifyOnStatuses`, a Teams 1:1 DM (Adaptive Card with request id, requester, project, environment, build, status, timestamps, duration, deep-link) is sent to the requester within ~10 seconds of status persistence.
- [ ] Notification failure of any kind (credential, network, resolution, missing bot install) does not affect deployment status updates or event publishing; failures are logged.
- [ ] A user without the bot installed simply doesn't receive the DM — logged, deployment unaffected.
- [x] `TeamsNotification.Enabled = false` registers the no-op sink with no side effects (unit-tested).
- [x] Unit + integration tests cover happy path, retry, swallowed-failure, and no-op-when-disabled.
- [ ] No changes to the public deployment API surface; no secret material committed. *(Holds in the diff — final check at merge.)*
- *(Unchecked items need manual verification in ut2 once PR #669 merges and the bot is installed.)*

## Unknowns Register

| # | Unknown | Blocking? | Status |
|---|---|---|---|
| U1 | AAD app + Bot Service provisioning | — | ✅ Resolved — registrations created, configured for ut2 |
| U2 | Graph `Chat.Create` tenant policy | — | ✅ Resolved — Graph chat approach abandoned; Bot Framework proactive messaging requires no Graph chat permissions |
| U3 | Teams app distribution — tenant-wide catalog vs sideload | Yes for go-live | Open — confirm with platform team |
| U4 | Requester identity mapping | — | ✅ Resolved — `UserName` is usually the email claim; resolved via `AzureEntraSearcher` |
| U5 | Behaviour for M2M/machine-account requesters (`UserName` = client id) | No — resolution fails safe (skip + log) | Open — decide if machine requests should notify anyone |
| U6 | Adaptive Card schema version | — | ✅ Resolved — 1.4 in use |

---

_Original ask: #385 (@benhegartysefe). Graph-based rewrite 2026-04-30; rewritten by @claude on 2026-08-02 to match the as-built Bot Framework implementation in #669 (per @slafeer-sefe's findings); checklist updated 2026-08-02 after the gap-closure work was pushed to #669._

slafeer-sefe:


hegsie:
if a job starts many at once is the user going to get flooded with message or just a grouped update?

hegsie:
Good question — here's the actual behaviour as implemented.

**One DM per deployment request, never per component.** A request with 50 components produces exactly one card, sent once the component loop finishes and the request reaches its terminal status (`PendingRequestProcessor.cs:319`). Per-component results, progress and intermediate transitions are silent.

**Nothing is sent on start.** Only terminal statuses notify, and only those listed in `TeamsNotification.NotifyOnStatuses` (default `Completed,Failed,Errored`). `Running`, `WaitingConfirmation` etc. don't DM unless you add them to that list.

**There is no cross-request grouping**, so if one user submits N separate requests they get N cards, each arriving as its own request finishes. Two cases where those can bunch up:

- A user batch-submitting several requests — the cards are spaced by however long each deployment actually runs, so in practice they trickle rather than burst.
- A monitor restart or a bulk cancel/abandon sweep in `DeploymentRequestStateProcessor` — it walks the requests it's switching and fires one DM per request whose transition it won (`DeploymentRequestStateProcessor.cs:303`). That's the realistic burst case. Each card still goes only to that request's own requester, so the blast radius is per-user, but someone who had 10 requests in flight when the Monitor restarted would get 10 "Errored" cards close together.

Mitigation available today with no code change: `NotifyOnStatuses` is per-environment config, so a noisy tier can be set to `Failed,Errored` only.

If you want real grouping (a debounce window keyed on requester, collapsing into a digest card), that's a design change with its own state to hold — I'd rather do it as a follow-up on #657 than widen this PR. Say the word and I'll raise the issue.

---

hegsie:
if we have a lot of requests at once does the user get flooded rather than just get a grouped message

hegsie:
could this also replace the send email in the bundle? which is just a hard coded powershell script

hegsie:
Short answer to the direct question: **yes — flooded, one card per request. There is no grouping today.**

(Fuller breakdown in [my comment above](https://github.com/sefe/dorc/pull/669#issuecomment-5523490350), posted just before this — the one thing that *is* grouped is components: a 50-component request sends one card, not 50.)

Two ways to fix it, if you want it fixed:

1. **Group the bulk sweeps** — a monitor restart or mass cancel switches many requests in one pass and currently fires one DM per request. That batch is already in hand in `DeploymentRequestStateProcessor`, so it can collapse to one card per requester ("3 of your requests were Errored") with no timers or new state. Contained change, I can put it on this PR.
2. **Group requests completing over time** — needs a debounce window per requester and delayed delivery, which trades away the "DM lands within ~10s of completion" behaviour. Bigger, and I'd rather it be a follow-up on #657 than widen this PR.

My recommendation is (1) here and (2) as a separate issue, since (1) covers the actual burst case. Tell me which and I'll do it.

---

hegsie:
Partly — worth doing, but it can't be a straight swap. What's actually there:

**DOrc itself sends no email.** There's no SMTP client anywhere in the repo. What survives is dead legacy in Dorc.Monitor: `Template/email.vm` (embedded resource), `TemplateEngine.cs` and the `NVelocity` package reference — all three have zero callers. The live mailout is bundle-side PowerShell, and DOrc's only involvement is feeding it the `EnvOwnerEmails` deployment property (`PendingRequestProcessor.SetUpEnvOwnerEmailAsProperty`, `VariableScopeOptionsResolver.cs:147`).

**Where the sink replaces it cleanly:** notifying the person who asked for the deployment. That's strictly better than a bundle script — no per-bundle copy to maintain, no SMTP config, no script ordering, and a deep link straight to the result.

**Where it doesn't, as built:**

- *Audience.* The sink DMs `request.UserName` — the requester. The bundle email goes to `EnvOwnerEmails` — the environment owner(s), a different set of people and often more than one.
- *Reachability.* A proactive DM only lands if the recipient has the bot installed in Teams. Distribution lists, shared mailboxes and external addresses can't receive one at all — the sink already skips entries that resolve to a group rather than a user. Email has no such prerequisite, so a Teams-only replacement silently drops those recipients.
- *Content depth.* The old template carried per-component results and logs inline; the card is a summary plus a link. Fine if people are happy to click through, a regression if they relied on the log in the mail.

**To actually retire the bundle script** you'd need: recipient resolution from `EnvOwnerEmails` (not just the requester), per-recipient dispatch, and a decision on unreachable recipients — skip them, or keep an email path as fallback so DLs still get told. That's a sensible v2 and I'd rather scope it on #657 than bolt it onto this PR.

The dead `email.vm` / `TemplateEngine` / NVelocity trio can go whenever — unrelated to this change, so I've left it alone here. Happy to raise both as follow-ups if you want.

---

hegsie:
APP ID is missing need this from the APP reg please 
image

hegsie:
I can't supply that one — the app id is a value from your Azure tenant, not something held in the repo or that I have access to. Two places to get it:

1. **Azure Portal** → App registrations → the DOrc notification bot → Overview → **Application (client) ID**.
2. Quicker, if you have it to hand: it's already configured as `TeamsNotification:BotAppId` (MSI parameter `TEAMS.BOT.APP.ID`) on the ut2 Monitor, so you can read it straight out of that environment's config.

It isn't a secret — unlike `TEAMS.BOT.APP.PASSWORD` — so it's fine to paste into the manifest.

What did bite you is that the placeholder appears in **two** places in `manifest.json` (`id` and `bots[0].botId`) and substituting only one leaves the catalog rejecting the package. Pushed `1c2fbb1` to stop that recurring:

- **`teams-app/package.ps1`** — takes `-BotAppId`, validates it parses as a GUID, substitutes both occurrences into a staging copy (so the checked-in manifest keeps its placeholder and can't be committed half-substituted), and writes the flat, BOM-free zip Teams wants:

  ```powershell
  cd teams-app
  ./package.ps1 -BotAppId <application-client-id-guid>
  ```

- Setup doc now spells out which value it is and where to find it, and keeps the manual zip route documented for anyone who'd rather do it by hand.

Note I couldn't execute the script here — there's no PowerShell in this container — so it's reviewed but not run. Worth a dry run on your side before relying on it.

---

hegsie:
Merged current `main` (`c64d7b4`) in — the branch was 88 commits behind and `mergeable_state` had gone to `dirty`. Now rebased onto current main and building.

**Conflicts, both in this branch's territory:**

- `DeploymentRequestStateProcessor` — main threaded `ILoggerFactory` through the constructor while this branch added `IDeploymentNotificationSink`. Kept both; each is genuinely used (the sink by `FireNotification`, the factory when handing off to `PendingRequestProcessor.Execute`).
- `.gitignore` — kept both entries.

**Worth flagging separately: `Dorc.Monitor.Tests` does not compile on `main` at `c64d7b4`.** `ILoggerFactory` was added to `PendingRequestProcessor.Execute` and to `DeploymentRequestStateProcessor`'s constructor without updating the call sites in `PendingRequestProcessorStopOnFailureTests` (7 calls) and `DeploymentRequestStateProcessorTests` (2 ctor + 4 matchers) — 20 × CS7036. I verified this against a clean checkout of `main`, not just my merge result, so it predates this branch.

I've repaired those call sites here because otherwise this PR can't build — real calls take `NullLoggerFactory.Instance`, NSubstitute matchers take `Arg.Any<ILoggerFactory>()`. That does mean this PR now carries a fix that logically belongs to whichever change threaded the logger factory through. Happy to split it into its own PR against `main` if you'd rather land it independently — it would unblock `main`'s test build straight away.

Local after the merge: `Dorc.Monitor` builds clean, `Dorc.Monitor.IntegrationTests` builds clean, `Dorc.Monitor.Tests` **110 passed / 0 failed / 3 skipped**.

---

hegsie:
Implemented the bulk-sweep grouping from the flood question above (`52c385a`, pushed as `d8f245f` after merging the newer branch head in).

**What changed:** a monitor restart or mass cancel switches many requests in one pass and used to fire one DM per request — ten in-flight requests meant ten cards back to back. Both sweeps already hold the whole batch, so grouping needed no timers and no new state:

- `IDeploymentNotificationSink` gains `NotifyRequestsCompletedAsync` for a set of requests that reached the same status together.
- `TeamsBotNotificationSink` groups the batch by requester (case-insensitively), resolves each recipient once, sends one card. Recipients are isolated — an unresolvable user or a Bot Connector failure for one person doesn't stop the others being told. A batch of one still sends the normal single-request card rather than degrading to a summary.
- `DeploymentCompletionCardBuilder.BuildBatch` renders the summary — status, count, and the request list capped at 10 with an "…and N more" line — deep-linking to `/monitor-requests`, since a batch can't point at one result.
- `DeploymentRequestStateProcessor` fires once per sweep at both sites (stale `Requesting` cleanup, and `SwitchRequestsStatus` for cancel/abandon), still only for the transitions this instance won, so a concurrent monitor can't DM the same requester twice.

**Unchanged:** requests completing over time still send one card each. Collapsing those needs a debounce window per requester and costs prompt delivery — left for #657 as discussed.

Tests: the five existing sweep tests were rewritten to assert batch semantics (they failed against the new behaviour, which is the point), plus 11 new sink tests covering same-requester grouping, multi-requester fan-out, case-insensitive grouping, batch-of-one, per-recipient failure isolation, empty/blank usernames and the status filter, and 6 card tests covering truncation and the deep link. **127 passed / 0 failed / 3 skipped** locally.

---

PR #683, opened 2026-05-08
add script to fix issue with pr environments issecure values

slafeer-sefe:


src/Dorc.Database/Scripts/Post-Deployment/UpdateEnvironmentSecure.sql
kirill-suspitsin-ext-sefe, line 1: No DB changes necessary, only UI changes to show current DB flags for existing environments

PR #706, opened 2026-05-28
HLPS-DRAFT: Split Dorc.Api into Linux-compatible + Windows-only worker (#423)

### LINKED ISSUE #423: Split the Dorc API into 2 different API's
1. first API should contain all code that is able to be moved to a linux environment
2. second API should be the rest of the code/ bare minimum code thats needed for running on windows

there should be a way of interfacing from the first API to the second when required

benhegartysefe:
## Summary

- Drafts the HLPS for issue #423 (split `Dorc.Api` into a Linux-compatible primary and a Windows-only worker), scoped to the split only — explicitly excluding the bulk class-rename that bundled into the superseded #424.
- Carries the five analysis docs from #424 over to `docs/api-split/research/` so the investigation isn't lost.
- Flags three blocking unknowns (worker process topology, AD migration direction, inter-API auth propagation) that need architecture-owner sign-off before an IS can be drafted.

This PR is **documentation-only** — no code changes, no installer changes. It is the first checkpoint of the HLPS → IS → SPEC process for #423.

## Why this exists

#424 was closed unmerged ([rationale](https://github.com/sefe/dorc/pull/424#issuecomment-4563797132)) because it bundled the structural split with:
1. A bulk class-rename pass that misread CLAUDE.md's naming rule as a banned-words blacklist (producing grab-bag names: `Properties`, `Requests`, `Operations`).
2. Wide drift after daemons-modernisation (#649) landed on `main` — ~340 silently semantic-conflicted files.
3. Three unresolved architectural decisions baked into code without sign-off.

This HLPS unwinds (1) and (3) and asks the architecture owner to resolve the unknowns before any code is written.

## Status

`DRAFT` — per CLAUDE.md, HLPS progresses DRAFT → IN REVIEW → REVISION → APPROVED. Only the adversarial review panel can approve.

## Test plan

- [ ] Architecture owner reviews and answers U-1 (worker topology), U-2 (AD migration direction), U-3 (auth propagation).
- [ ] Adversarial review panel evaluates clarity, ordering, completeness of the HLPS (per CLAUDE.md review scope for plans).
- [ ] On approval, draft `docs/api-split/IS-api-split.md` ordering atomic steps; second checkpoint there.


benhegartysefe:
Superseded by a five-PR stack, rebased onto current `main`. Closing this in favour of:

| PR | Content | Base |
|----|---------|------|
| #863 | HLPS + research docs | `main` |
| #864 | Implementation Sequence | #863 |
| #865 | SPEC-S-001 | #864 |
| #866 | AD→Graph production code + tests | #865 |
| #867 | Linux build CI gate | #866 |

Merge in order — #867 can only pass once #866 has removed the Windows-only refs.

**Why split:** this PR had grown from an HLPS draft into planning docs *plus* production code, tests and a CI workflow across 15 commits, while its description still said "documentation-only". It was also conflicting with `main`. The split follows the CLAUDE.md planning checkpoints, so each planning artifact is independently approvable and the only behaviourally risky change (#866) is reviewable on its own.

**Two things the rebase and review surfaced, both now fixed:**

`main` landed LDAP-injection hardening on `ActiveDirectorySearcher`. `AzureEntraSearcher` carried an identical copy of the broken regex that the fix never touched — so deleting the fixed class and keeping the other would have reverted the fix in effect, with **no git conflict**, because the two regexes live in different files. Carried forward in #866. Note `main`'s own copy still anchors with `$`, which in .NET matches before a trailing newline.

An adversarial review panel then found a **CRITICAL** defect: WinAuth passes a bare sAMAccountName to `GetSidsForUser`, which Graph rejects with 400 as a path segment — every environment/property endpoint would have 500'd under WinAuth. Plus silent truncation of group membership at 100 (paged `transitiveMemberOf`), an unreachable SID fallback (Graph returns 400, not 404), and a name-resolution path that could bind a caller to another principal's group claims. All fixed in #866, with the full list in its description.

Mutation testing showed **six of eight injected defects passed the original test suite**, including an authorization bypass; the harness was rebuilt to assert outbound Graph requests and all eight now fail. Separately, the CI gate's build steps did not actually fail on Windows-only code — CA1416 is a warning unless promoted — so #867 rebuilds the enforcement.

Everything verified on Linux with .NET SDK 8.0.424: all gated projects plus `Dorc.Monitor` and the five `Tools.*` CLIs build clean; `Dorc.Core.Tests` 164/164 and `Dorc.Api.Tests` 243/243.

---

PR #710, opened 2026-05-28
S-009: Log-injection carve-outs (BundledRequests, MakeLikeProd, RequestService)

benhegartysefe:
## Summary

- IS step S-009 from the [API split work](https://github.com/sefe/dorc/pull/706). Closes HLPS SC-8b for the controllers on the non-worker move path.
- Adds `Dorc.Core/LogSanitizer.cs` — strips ASCII control chars from values that flow into log statements so user-controlled input cannot forge log lines or smuggle terminal escape sequences. Caps length at 512.
- Reworks the four sites where `$"…{userInput}…"` was being passed as the structured-logging template (CWE-117 log injection):
  - `BundledRequestsController` — 5 sites
  - `RequestService` — 3 sites
  - `MakeLikeProdController` — 1 site (TargetEnv sanitisation)
- `ResetAppPasswordController` is *not* touched here — its move to the Windows worker (S-006) handles the sanitisation as part of that move. SPEC documents the deferral.

Independent of PR #706 per the IS's multi-PR plan — can merge in any order.

## Test plan

- [x] `Dorc.Core.Tests/LogSanitizerTests.cs` — 11 new tests (null, empty, printable ASCII, non-ASCII, `\r\n`, `\t`, `\0`, terminal escape, DEL, 512-char cap, attacker multi-line payload). All pass.
- [x] `Dorc.Core.Tests`: 131 / 131 pass (11 new + 120 existing)
- [x] `Dorc.Api.Tests`: 190 / 190 pass (zero regressions)
- [x] `Dorc.Api`, `Dorc.Core`, `Dorc.PersistentData` build clean
- [ ] SAST re-scan against this branch should report zero log-injection findings on the three modified files. ResetAppPasswordController findings (if still present) are expected per the deferral.

### DIFF
--- a/docs/api-split/SPEC-S-009-log-injection.md
+++ b/docs/api-split/SPEC-S-009-log-injection.md
@@ -1,0 +1,87 @@
+---
+name: SPEC-S-009 — Log-injection carve-outs for primary API controllers
+description: JIT Specification for S-009 — adds a small sanitisation helper and reworks four log sites in the primary API so user-controlled values cannot inject newlines or break structured-logging templates. Closes HLPS SC-8b for everything except the ResetAppPasswordController log site (S-006 territory).
+type: spec
+status: APPROVED
+---
+
+# SPEC-S-009 — Log-injection carve-outs for primary API controllers
+
+| Field       | Value                                              |
+|-------------|----------------------------------------------------|
+| **Status**  | APPROVED (auto-pilot per user direction)           |
+| **Step**    | S-009                                              |
+| **Author**  | Agent                                              |
+| **Date**    | 2026-05-28                                         |
+| **IS**      | [IS-api-split.md](IS-api-split.md) (APPROVED)      |
+| **HLPS**    | [HLPS-api-split.md](HLPS-api-split.md) (APPROVED)  |
+
+---
+
+## 1. Context
+
+The IS lists four controllers (`ResetAppPasswordController`, `BundledRequestsController`, `MakeLikeProdController`, and the former `Deployment/Requests.cs`) carrying log-injection findings from the PR #424 SAST. The PR #424 file paths and line numbers are stale; current `main` (anchor `f0fbf917`) has equivalent sites at:
+
+- `src/Dorc.Api/Controllers/BundledRequestsController.cs` lines 51–53, 75–77, 176–178 — `_logger.LogError(ex, error)` where `error` is a string built via `$`-interpolation / concatenation around user input (`projectNames`, `bundleName`, `id`). The interpolated string is then used as the message template, which (a) breaks structured-logging caching, (b) lets user input introduce `{}` tokens that corrupt downstream args, and (c) allows newline/control characters to forge log lines.
+- `src/Dorc.Api/Services/RequestService.cs` lines 29 and 52–53 — same antipattern: `_log.LogError($"...{userInput}...")`.
+- `src/Dorc.Api/Controllers/MakeLikeProdController.cs` line 223 — already uses structured templates (`{TargetEnv}`, `{@Request}`), but the user-controlled `TargetEnv` value can still contain newlines that appear literally in formatted output. `{@Request}` JSON-serialises the whole DTO; out of scope here.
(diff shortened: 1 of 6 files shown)


benhegartysefe:
**Rebased onto current `main`** — this branch was 214 commits behind. Your commit (`af4271c`) cherry-picked with no conflicts.

**Base deliberately left as `main`.** The rest of the `#423` series has been folded into a stack (#863 → #864 → #865 → #866 → #867 → #711 → #712 → #714 → #713), but this PR stays independent, as its own description says: *"Independent of PR #706 per the IS's multi-PR plan — can merge in any order."* Nothing in the log-injection work depends on the API split, so stacking it would only delay it behind nine other PRs.

`Dorc.Api` builds clean and `Dorc.Core.Tests` passes on current `main`.

Note the test counts in the description are stale relative to `main` (it cites `Dorc.Core.Tests` 131/131 and `Dorc.Api.Tests` 190/190; both suites have grown since). Worth refreshing before merge.

---

benhegartysefe:
## Review: the sanitiser had real gaps, fixed

`LogSanitizer` kept everything at or above U+0020 except DEL, so it stripped CR/LF/TAB/ESC but let every **non-ASCII** line-breaker and the entire **C1 control block** through. I compiled the old and new implementations side by side and ran them:

| Input | Before | After |
|---|---|---|
| `
` NEL | **survived** | stripped |
| `` CSI | **survived** | stripped |
| `` PAD | **survived** | stripped |
| `
` LS | **survived** | stripped |
| `
` PS | **survived** | stripped |
| `` RLO | **survived** | stripped |

Why each matters: U+0085/U+2028/U+2029 are line terminators to .NET `Regex` in multiline mode and to every JavaScript-based log viewer, so log forging still worked with a different byte. U+009B is the single-character CSI — blocking ESC (0x1B) doesn't stop ANSI injection into an operator's `tail -f`. U+202E visually reverses the rest of an audit line.

`char.IsControl` covers C0, DEL **and** C1 in one move; separators and the Unicode `Format` category are handled explicitly. Real names still pass through untouched — the check keys off Unicode category, not "is it ASCII", so `Müller`, `Łukasz` and `José` are unaffected.

Also fixed truncation splitting a surrogate pair at the 512 cap, leaving a lone high surrogate that serialises downstream as U+FFFD.

**On the test suite:** answering the obvious question — a mutant returning its input unchanged *was* caught by the existing tests. But a mutant stripping only CR/LF/TAB/NUL/ESC/DEL passed everything, which is precisely why the gap shipped. `NonAscii_SurvivesUnchanged` actively asserted the behaviour that caused it. Six escape-class cases added; all six fail against the previous implementation, confirmed by reverting and re-running. 157/157 now.

## Two things I did not change

**The `RequestService` sanitisation is defeated one frame up.** `RequestService.cs:54` logs the sanitised project name, then line 55 throws `new Exception($"...'{projectName}'")` carrying the **raw** value, and `RequestController.cs:624` does `_log.LogError(e.Message)`. Net effect of the change on that path is zero: `Project = "X\r\n2026-08-17 12:00:00 [INF] Deployment 999 approved by admin"` still forges a line. Same for the raw `BuildUrl` and `ValidationResult` in the exceptions at `:32` and `:44`. Worth noting `e.Message` is passed as the *template*, so a `{` in a project name is parsed by Serilog as a property token.

**`MakeLikeProdController` sanitises the one value that was already safe.** The diff touches only the catch block, where `{@Request}` destructures the DTO — and Serilog's structure formatter already JSON-escapes strings inside destructured objects. Meanwhile `BundleName`, `Project` and `ProjectName` are rendered verbatim on the **success** path at `:170`, `:183`, `:192`, `:211`, all from `[FromBody]` or deserialised bundle payloads. The success path is the one an attacker can trigger reliably.

Both are in this PR's own files, so they're in scope — I left them because fixing them properly means deciding whether to sanitise at the boundary or stop passing raw `e.Message` as a template, and the same pattern recurs in `DefaultExceptionHandler.cs:46`, which is the single highest-value fix in the codebase for this class of bug and is outside this diff. Related un-sanitised sites, for scope-setting rather than action: `RequestController.cs:602`, `TerraformController.cs:131,194`, `PropertiesService.cs:202`, `OnePasswordSecretsReader.cs:32`. A Roslyn analyzer banning interpolated strings as the first argument to `ILogger.Log*` would be cheaper than auditing each.

Minor: `LogSanitizerTests` embeds a literal 0x7F byte in the source; `"ab"` is less fragile.

Rebased onto current `main` (was 214 commits behind), base deliberately left as `main` per your note that this is independent. Builds clean.

---

PR #712, opened 2026-05-28
S-003: IWindowsWorkerClient contract + null-impl + worker-absence detection

benhegartysefe:
## Summary

- IS step S-003 from the [API split work](https://github.com/sefe/dorc/pull/706). Establishes the seam every later worker-move step (S-004 / S-005 / S-006) extends.
- New `IWindowsWorkerClient` interface (empty marker for now; later steps add methods).
- Two implementations: real `HttpWindowsWorkerClient` (typed HttpClient with `WorkerKeyDelegatingHandler` injecting `X-Worker-Key`), and `WorkerUnavailableClient` (null impl that throws `WorkerUnavailableException` → global filter renders `503 {"error":"windows_worker_unavailable","endpoint":"…"}`).
- New `WindowsWorker:Enabled` config flag selects the implementation. Defaults to `false` so existing Windows installs behave unchanged at this commit (flag flips `true` in S-008 when the installer wires the worker).
- Resolves HLPS U-6 (URL discovery — config-file), U-7 (contract via shared DTOs in `Dorc.ApiModel` — to be exercised by later steps), U-11 (worker-absence detection — explicit config flag).

## Test plan

- [x] `Dorc.Api.Tests/WindowsWorkerClientTests.cs` — 4 new tests covering the delegating handler's header injection (with and without configured secret) and the global exception filter's behaviour (translate + ignore).
- [x] `Dorc.Api.Tests`: 194 / 194 pass (4 new + 190 existing).
- [x] `Dorc.Api` builds clean.

Independent of PR #706 / #710 / #711 per IS multi-PR plan. S-004 / S-005 / S-006 stack on this once it merges.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #711, opened 2026-05-28
S-002: Dorc.Api.WindowsWorker scaffold + loopback host + shared-secret auth

benhegartysefe:
## Summary

- IS step S-002 from the [API split work](https://github.com/sefe/dorc/pull/706). Creates the empty Windows worker project — endpoint-empty for Windows operations; S-004 / S-005 / S-006 add the controllers.
- New project `src/Dorc.Api.WindowsWorker` (`net8.0-windows`, `Microsoft.NET.Sdk.Web`). Kestrel binds 127.0.0.1 only (HLPS D-1).
- `WorkerKey` authentication scheme reads `X-Worker-Key` header, validates constant-time vs `WindowsWorker:SharedKey`. Missing/wrong → `401` + body `{"error":"worker_key_invalid"}` (HLPS D-3).
- `FromPrimary` authorization policy gates controller routes; `/health` is intentionally unauthenticated (loopback bind is the security boundary).
- Independent of PR #706 per IS multi-PR plan.

## Test plan

- [x] `Dorc.Api.WindowsWorker.Tests`: 7 / 7 pass (5 handler unit tests + 2 host smoke tests on `/health`)
- [x] Worker project + test project build on Windows
- [x] Both projects added to `src/Dorc.sln`
- [ ] Full-pipeline coverage of protected controllers will land in S-004/S-005/S-006 when the actual Windows endpoints are added
- [ ] Worker is intentionally NOT yet built on Linux (TargetFramework `net8.0-windows`); the Linux build CI gate from S-001 only checks the primary's compile graph

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #713, opened 2026-05-28
S-007: Remove Negotiate / WinAuth scheme from primary API

benhegartysefe:
## Summary

- IS step S-007 from the [API split work](https://github.com/sefe/dorc/pull/706). Removes the last Windows-only auth dependency from the primary's compile graph (HLPS Scope E).
- Deletes `ClaimsTransformer`, `WinAuthClaimsPrincipalReader`, `WinAuthLoggingMiddleware`.
- Simplifies `Program.cs`: drops Negotiate using, removes `ConfigureWinAuth` / `ConfigureBoth`, replaces the auth-scheme `switch` with a startup guard that throws with a clear migration message if `AuthenticationScheme` is `WinAuth` or `Both`.
- Collapses `ClaimsPrincipalReaderFactory` to its one remaining branch (OAuth). Name not changed (HLPS C-2 — separate naming pass).

**Stacked on #706 (S-001).** Base branch is `feat/423-api-split-hlps`, not main. Cannot merge until #706 merges; PR base will auto-retarget to main when that happens.

**Customer impact:** any install with `AppSettings:AuthenticationScheme` set to `WinAuth` or `OAuth+WinAuth` (current default in `appsettings.json`) will fail to start with a migration message. Fixing the default in `appsettings.json` is S-008's installer responsibility.

## Test plan

- [x] `Dorc.Api`, `Dorc.Core`, `Dorc.PersistentData`, `Dorc.Api.Tests` build clean
- [x] `Dorc.Api.Tests`: 190 / 190 pass (no regressions)
- [x] `git grep "WinAuth\|Negotiate" src/Dorc.Api` returns hits only in the new startup-guard message and the legacy `appsettings.json` default (latter is S-008's fix)


benhegartysefe:
**Rebased and retargeted.** This PR was stacked on `feat/423-api-split-hlps` (PR #706), which has been closed and replaced by a five-PR stack. As written, this PR's note that "PR base will auto-retarget to main when #706 merges" would never have fired — it was orphaned on a base that will never merge.

New base: `feat/423-s004-registry-move` (#714), which now sits on top of the S-001 stack:

`main` → #863 HLPS → #864 IS → #865 SPEC → #866 S-001 → #867 Linux gate → #711 S-002 → #712 S-003 → #714 S-004 → **this PR**

Your single commit (`9cb5ad2`) cherry-picked onto the new base with **no conflicts**; the branch was force-pushed with the rebased commit. `Dorc.Api` builds clean and `Dorc.Api.Tests` passes 247/247 on the new base.

One interaction worth flagging: this PR deletes `WinAuthClaimsPrincipalReader`, which is the caller behind a CRITICAL defect fixed in #866 — `GetSidsForUser` was being handed a bare sAMAccountName that Graph rejects with 400 as a path segment. Removing WinAuth does **not** make that fix redundant: `ClaimsPrincipalReaderFactory.GetSidsForUser` still calls `_adUserGroupsReader.GetSidsForUser(GetUserLogin(user))` when `IsUseAdSidsForAccessControl` is set, and under OAuth `GetUserLogin` returns a sAMAccountName or email. The fix in #866 is still load-bearing after this PR.

---

benhegartysefe:
## Review: one CRITICAL, fixed

**The startup guard this PR adds would have bricked every install.** `appsettings.json` ships `"AuthenticationScheme": "OAuth+WinAuth"` — which is exactly `ConfigAuthScheme.Both`, the value the guard throws on. Verified by reading both files on this branch. Worse, `RequestApi.wxs:49` and `:141` write that same path from the `[DORC.AUTHENTICATION.SCHEME]` MSI property on every install and upgrade, so any customer whose saved property is `WinAuth` or `OAuth+WinAuth` gets a failed upgrade. And `configBuilder` loads only `appsettings.json` — no environment overrides — so there was no escape hatch.

Fixed here:
- Default flipped to `"OAuth"`.
- **Guard inverted to accept only `OAuth`** (case-insensitive). The original rejected two known-bad literals and let everything else through, so `null`, `""` or `"winauth"` configured OAuth with a null Authority and failed on the *first request* instead of at startup — the exact outcome the guard exists to prevent.
- The throw happens before `builder.Build()` and outside any handler. Under ANCM in-process hosting that surfaces as a bare 500.30, with the carefully-worded migration message nowhere an operator looks. Now written through Serilog and flushed first.

**Still needs your action:** MSI upgrades keep writing the saved property. S-008 needs a custom action rewriting `WinAuth`/`OAuth+WinAuth` to `OAuth`, or the guard will block upgrades regardless of the new default.

## Also fixed: the Negotiate removal was incomplete

`ContextExtensions.cs` still returned `NegotiateDefaults.AuthenticationScheme` for any non-Bearer request, and was the sole remaining reason for the `Microsoft.AspNetCore.Authentication.Negotiate` package reference. Both had zero callers once `ClaimsTransformer` and `WinAuthLoggingMiddleware` were deleted. Removed — leaving an "assume Negotiate" default in a PR whose purpose is removing Negotiate invites its reintroduction.

## Two decisions I did not make for you

**`ClaimsPrincipalReaderFactory` is now dead code, and `IsUseAdSidsForAccessControl` is a silent no-op.** `ConfigureBoth` was its only wiring; `IClaimsPrincipalReader` now resolves solely to `OAuthClaimsPrincipalReader`. The flag is read only from inside the unreachable class. Sites running with it `true` previously got AD SIDs and now get Entra oids/pids, so `AccessControl` rows keyed on legacy AD SIDs stop matching — fails closed, but silently. Its class comment claimed consumers exist; I corrected the comment rather than deleting the class, because whether that flag should keep working is your call. Note that even if you re-register it, its AD branch resolves names via `GetUserName`/`GetUserLogin`, which under OAuth yield a display name or email that `ResolveUserIdFromName` won't match — so "just register it" doesn't restore the behaviour either.

**`AppSettings:ActiveDirectoryRoles` is now consumed by nothing** but the installer still populates it from `[AD.ROLEPROVIDER.GROUP.ADMIN]`/`[…POWERUSER]`. `ClaimsTransformer` was its only reader, while `User.IsInRole("Admin")`/`("PowerUser")` remain load-bearing across `RefDataController`, `RefDataPermissionController`, `DeploymentsHub`, `RolePrivilegesChecker` and `SecurityObjectFilter` — now resolving solely from the JWT `role` claim. An operator who upgrades and sets the AD admin group as always gets 403s everywhere if the Entra app registration doesn't emit `role`. That prerequisite belongs in the S-010 migration guide.

One more worth confirming before merge: `RequireAuthorization(apiScopeAuthorizationPolicy)` is now unconditional, where previously it applied only under scheme `OAuth` — i.e. not under the shipped `OAuth+WinAuth` default. Any M2M client provisioned before the global scope existed will start getting 403 on every endpoint. `[AllowAnonymous]` on `ApiConfigController` is still honoured, so the UI bootstrap is fine.

`Dorc.Api` builds clean; `Dorc.Api.Tests` 251/251.

---

benhegartysefe:
**Moved to draft.** Code-complete and green — not shippable, for a reason bigger than merge ordering.

The startup guard here rejects `AuthenticationScheme` values of `WinAuth` or `OAuth+WinAuth`. `RequestApi.wxs` writes that setting from `[DORC.AUTHENTICATION.SCHEME]` on **every install and upgrade** (lines 49 and 141), and also configures IIS with `WindowsAuthentication="yes"` / `AuthenticationProviders="Negotiate,NTLM"`.

**No step in the plan owns `RequestApi.wxs`.** Scope E is code-only; S-008 only *adds* the worker component. So as currently planned, this PR ships an API that will not start after an upgrade — and SC-7, the upgrade-safety criterion, passes anyway because it's scoped to the new worker component. That's a scope hole, not just sequencing. It's recorded as R5-13 in the HLPS (#863, Appendix C) and needs an owning step before this leaves draft.

Flipping the shipped `appsettings.json` default to `OAuth` (done in this PR) is necessary but not sufficient — the MSI overwrites it.

---

PR #735, opened 2026-06-12
fix(web): guard request-status-card against undefined deployRequest

benhegartysefe:
## Problem

The Monitor result page throws a console error on load:

```
Uncaught (in promise) TypeError: can't access property "Log", this.deployRequest is undefined
    render index-...js:16840
```

## Root cause — a render race

`page-monitor-result.ts` kicks off three parallel subscriptions (request status, result statuses, attempts) and flips `loading` to `false` in **each** subscription's `complete` callback. The result-statuses and attempts calls routinely finish before the request call — visible in the console ordering (`done loading result Statuses` / `done loading attempts` arrive **before** `done loading request`). So `<request-status-card>` renders while `deployRequest` is still `undefined`.

`request-status-card.ts` then read `this.deployRequest.Log` (line 158) without optional chaining — the one access in the file that didn't use `?.` — and threw, aborting the render. The `<request-controls>` bindings (lines 308–321) had the same unguarded pattern and would have thrown next once line 158 was fixed.

## Fix

Add optional chaining to every `deployRequest` access in the render path, matching the convention already used throughout the rest of the file. The card now renders harmlessly until the request data arrives and the component self-heals on the next update.

- Cosmetic-but-noisy: the card eventually recovered, but the exception spammed the console and aborted a render pass on every load.
- `npx tsc --noEmit` is clean.

## Scope

`request-status-card.ts` only — diff-local, no behavioural change once data is loaded.

### DIFF
--- a/src/dorc-web/src/components/request-status-card.ts
+++ b/src/dorc-web/src/components/request-status-card.ts
@@ -155,7 +155,7 @@ export class RequestStatusCard extends LitElement {
                 .state="${this.hubConnectionState}"
               ></connection-status-indicator>
             </td>
-            ${this.deployRequest.Log !== null && this.deployRequest.Log !== '0'
+            ${this.deployRequest?.Log != null && this.deployRequest?.Log !== '0'
               ? html` <td style="vertical-align: middle">
                   <vaadin-button
                     title="View Log"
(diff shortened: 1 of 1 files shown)


hegsie:
Refreshed this branch against current `main` (commit 2ea6592).

**Conflict resolution** — `main` had reformatted `request-status-card.ts` and added `vaadin-tooltip`/`aria-label` markup plus `dorcApiConfiguration` wiring. I kept all of that and re-applied this branch's intent on top: optional chaining on every `deployRequest` access in the render path (`Log` guard and all `<request-controls>` bindings).

**Additional fixes while re-basing:**
- Guarded the remaining unguarded `deployRequest` reads that would have thrown next: `openEnvironmentDetails`, `copyRawLog`, `viewLog` and the build-link lookup in the lifecycle hook. The `Raw Log` row and the environment-details button render unconditionally, so those handlers are reachable while `deployRequest` is still undefined.
- The build-link lookup used to run once in `firstUpdated`; with optional chaining alone it would have fired `refDataProjectsProjectNameGet({ projectName: '' })` on an undefined request and then never retried. It now runs from `willUpdate` when `deployRequest` changes, is skipped when there is no project, and is de-duplicated per project name so status polling doesn't re-issue it.

Note the original race described above is partly gone on current `main` (`loading` is now only cleared by the request-details subscription), but the guard still matters: a non-404 error path clears `loading` without setting `notFound`, so the card can still render with an undefined request.

**Validation** (`src/dorc-web`): `npm run build` (eslint + tsc + typecheck:tests + vite build) clean; `npx prettier --check` clean; new `tests/components/request-status-card-undefined-request.test.ts` — 4 tests passing in chromium, firefox and webkit — covers rendering with no `deployRequest`, no build-link lookup without a project, recovery once the request arrives, and no refetch while the project is unchanged.

Issue #345, opened 2025-10-08
Tools.RequestCLI.exe fails when run with certain parameters


dmitryrext:
Tools.RequestCLI.exe gives exception when executed with certain parameters. Please reach @dmitryrext  for example
```
08/10/2025 04:29:28 - =============================================================================
08/10/2025 04:29:29 - Error creating request
08/10/2025 04:29:29 - The JSON value could not be converted to Dorc.Core.ContentResponse. Path: $ | LineNumber: 0 | BytePositionInLine: 27.
04:29:29 I  Activity 'ShellCommand' failed
04:29:29 E  ApplicationException in step 'ShellCommand': Process exited with '1'
04:29:29 E  The workflow has faulted
04:29:29 E  System.ApplicationException: Process exited with '1'
   at Asci.ActiveBatch.JobSteps.ShellCommand.Asci.ActiveBatch.Workflow.Runtime.IActivityExecutionHandler.Execute(IActivityExecutionContext executionContext)
   at Asci.ActiveBatch.Workflow.Runtime.ActivityExecutionContext.ExecutionActivityCore(ActivityExecutionContext executionContext, IActivityExecutionHandler executionHandler)
   at Asci.ActiveBatch.Workflow.Runtime.ActivityExecutionContext.ExecuteActivity(Activity activity)
   at Asci.ActiveBatch.JobSteps.JobStepSequence.Asci.ActiveBatch.Workflow.Runtime.IActivityExecutionHandler.Execute(IActivityExecutionContext executionContext)
   at Asci.ActiveBatch.Workflow.Runtime.ActivityExecutionContext.ExecutionActivityCore(ActivityExecutionContext executionContext, IActivityExecutionHandler executionHandler)
   at Asci.ActiveBatch.Workflow.Runtime.ActivityExecutionContext.ExecuteActivity(Activity activity)
   at Asci.ActiveBatch.Workflow.Runtime.WorkflowInstance.Run()
```

**tech detail**
When DOrc responds with 400 bad request, CLI should handle that properly and show the response instead of failing with parsing error

slafeer-sefe:
create request with build folder that doesnt exist eg: testfolder123 and watch for response - Build validation failed instead of parsing error

other parts of the cli request can be tested in the same way

PR #751, opened 2026-06-24
Code-review & security-audit fixes: error handling, build-URI parsing, authorization, LDAP & encryption hardening (with tests)

benhegartysefe:
## Overview

This PR began as fixes for a code review of recent `main` changes and grew to include fixes for four issues surfaced by a deeper whole-codebase security audit. Every behavioural change is covered by tests. Commits are grouped below; each entry explains **what** changed and **why**.

Test status (run locally on .NET 8): `Dorc.Core.Tests` 136 ✓, `Dorc.Api.Tests` (ConfigValues + AccessControl + BundledRequests) ✓, `Tools.RequestCLI.Tests` 7 ✓.

---

## 1. Error reporting — `Dorc.Core/ApiCaller.cs`

**What:** Replaced `result.ErrorMessage = responseContent.Trim('"')` with a new `ExtractErrorMessage` helper that:
- unwraps a bare JSON string body (e.g. `BadRequest("Project name must be provided.")`), correctly decoding escapes;
- extracts `Message`/`message` from a JSON **object** body (e.g. `CopyEnvBuildResponseDto`, serialized exceptions);
- also extracts RFC-7807 ProblemDetails `detail`/`title` (ASP.NET model-validation responses);
- falls back to the raw body when it is not JSON or has no recognisable, non-empty message.

**Why:** The previous `Trim('"')` only handled bare JSON strings. For object error bodies it surfaced the entire raw JSON blob to CLI users (e.g. `Tools.DeployCopyEnvBuildCLI` printing `{"RequestIds":[],"Success":false,"Message":"…"}`), and every `StatusCode(500, exception)` regressed the same way. An empty `Message` previously produced a blank error; it now falls back to the raw body.

**Tests:** `Dorc.Core.Tests/ApiCallerErrorMessageTests.cs` — bare string, object `Message`/`message`, ProblemDetails `detail`/`title`, empty-message fallback, plain text, non-string JSON.

## 2. Build-URI parsing — `Tools.RequestCLI/CliArgs.cs`

**What:** `/builduri:` is now parsed by splitting on the **first** colon only (`IndexOf(':')` + `Substring`), preserving the full URL; empty values are rejected with a clear `ArgumentException`.

**Why:** The previous code rebuilt the URL from `buildUri[1] + ":" + buildUri[2]`, keeping only the first two colon-delimited segments — so `https://host:8080/path` was silently truncated to `https://host` (port and path lost). The earlier guard fixed the crash from issue #345 but not the truncation.

**Tests:** `Tools.RequestCLI.Tests` (new project) — port+path preserved, no-port preserved, whitespace trimmed, empty/whitespace rejected, plus scalar/collection argument parsing. `CliArgs` internals exposed via `InternalsVisibleTo`.

## 3. Bundled-requests delete authorization cleanup — `Dorc.Api/Controllers/BundledRequestsController.cs`

**What:** Extracted the triplicated `CanModifyProject → 403 "User does not have Modify rights on this Project"` guard (Create/Update/Delete) into a single private `EnsureCanModifyProject(long?)` helper. Behaviour preserved (a missing project id still yields 403 without invoking the privilege checker).

**Why:** Three verbatim copies of the same authorization check/message drift over time; one helper keeps them consistent.

## 4. Frontend cleanups — `dorc-web`

- **`components/log-dialog.ts`** — removed the never-set `--log-viewer-height` CSS variable; the loading container now uses the same `calc(85dvh - 90px)` as the rendered viewer (single source of truth for the height).
- **`pages/page-monitor-requests.ts`** — removed the dead no-op `connectedCallback` override (left behind when the keydown listener was removed); added `type="button"` to the request-id button so it can never act as an implicit form-submit.
- **`pages/page-project-envs.ts`** — aligned `.card-element` `min-height` to `110px` to match `environment-card`, so adjacent card types render at a consistent height.

---

### C1 — ConfigValues credential disclosure (CRITICAL) — `Dorc.Api/ConfigValuesController.cs`, `ConfigValuesPersistentSource.cs`, `IConfigValuesPersistentSource.cs`

**What:** `GET /ConfigValues?name=` now requires Admin (`IRolePrivilegesChecker.IsAdmin`) and never returns `Secure` values. Added `IConfigValuesPersistentSource.GetNonSecureConfigValue`, which returns `null` for a missing **or** `Secure` key; the controller uses it instead of `GetConfigValue`.

**Why:** The endpoint was `[Authorize]`-only and returned any config value — including `Secure` ones that `GetConfigValue` decrypts, e.g. `DORC_NonProdDeployPassword` (read internally by `ResetAppPasswordController`). Any authenticated user could retrieve the deploy account password. Internal server callers continue to use `GetConfigValue` directly.

**Tests:** `Dorc.Api.Tests/Controllers/ConfigValuesControllerTests.cs` — non-admin forbidden (store not consulted), admin+secure key → not found, admin+non-secure key → value.

### C2 — AccessControl IDOR / privilege escalation (CRITICAL) — `Dorc.Api/Controllers/AccessControlController.cs`

**What:** `Put` now resolves the target object from `Name` (via `GetSecurableObjects`, as `Get` already does) and uses **that** `ObjectId`, ignoring the client-supplied `ObjectId` (400 if `Name` does not resolve).

**Why:** The method authorized against body `Name` but wrote ACLs to the separate body `ObjectId`, never cross-checked. A caller could pass the `Name` of an object they own with the `ObjectId` of one they do not, and grant themselves Owner/Write/ReadSecrets on the latter.

**Tests:** `Dorc.Api.Tests/Controllers/AccessControlControllerTests.cs` — new: writes land on the resolved object not the spoofed id; unauthorized → 403; unresolved name → 400. Existing owner-count tests updated to stub `GetSecurableObjects`.

### H1 — LDAP injection / directory enumeration (HIGH) — `Dorc.Core/ActiveDirectorySearcher.cs`

**What:** Fixed the validation regex (the `'-_` fragment was an unintended character range 0x27–0x5F that permitted `* ( ) \`; hyphen is now a literal, extracted to `internal static IsValidSearchName`), and applied `EscapeLdapFilter` to the interpolated value in `GetUserData` and `GetSidsForUser` (defence in depth).

**Why:** `GetUserData` interpolated unvalidated input straight into the LDAP filter, allowing wildcard enumeration / filter injection. Reachable from the user-search endpoints.

**Tests:** `Dorc.Core.Tests/ActiveDirectorySearcherValidationTests.cs` — accepts realistic names, rejects LDAP metacharacters, and verifies `EscapeLdapFilter` escaping.

### H2 — Secure-array encryption data corruption (HIGH) — `Dorc.PersistentData/Sources/PropertyValuesPersistentSource.cs`

**What:** Replaced `.Select(s => Encrypt(propertyValue.Value))` (which ignored the element and re-encrypted the whole serialized array) with symmetric `EncryptArrayValue`/`DecryptArrayValue` helpers that encrypt/decrypt **each element**, and routed every secure-value decrypt site through an array-aware `DecryptStoredValue`.

**Why:** Secure arrays stored the ciphertext of the entire blob for every element, and the decrypt paths were not array-aware, so secure-array values were unrecoverable on both ends. **Note:** any secure-array data already written by the buggy code is corrupt and cannot be recovered by this change.

**Tests:** `Dorc.Core.Tests/PropertyValuesArrayEncryptionTests.cs` — per-element encryption, encrypt→decrypt round-trip, distinct ciphertexts. Internals exposed to `Dorc.Core.Tests` via source-level `InternalsVisibleTo` in `Properties/AssemblyInfo.cs` (the project sets `GenerateAssemblyInfo=False`, so the `.csproj` item is ineffective).

---

## Notes
- `.NET 8` SDK was used locally to build and run the test suites (`dotnet` is not part of the base web/CI image here).
- A full written security-audit report (4 CRITICAL, ~16 HIGH, plus MEDIUM/LOW across the whole codebase) was produced but is intentionally **not** included in this PR, as it contains concrete exploit detail. Available on request.

https://claude.ai/code/session_01Gg27w4VA4Rm6XpMWaN1eHf

### DIFF
--- a/src/Dorc.Api/ConfigValuesController.cs
+++ b/src/Dorc.Api/ConfigValuesController.cs
@@ -1,8 +1,7 @@
-using Dorc.Api.Interfaces;
-using Dorc.ApiModel;
-using Dorc.Core.Interfaces;
+using Dorc.PersistentData;
 using Dorc.PersistentData.Sources.Interfaces;
 using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 
 namespace Dorc.Api.Controllers
@@ -13,19 +12,31 @@ namespace Dorc.Api.Controllers
     public class ConfigValuesController : ControllerBase
     {
         private readonly IConfigValuesPersistentSource _configValuesPersistentSource;
-        public ConfigValuesController(IConfigValuesPersistentSource configValuesPersistentSource)
+        private readonly IRolePrivilegesChecker _rolePrivilegesChecker;
+
+        public ConfigValuesController(
+            IConfigValuesPersistentSource configValuesPersistentSource,
+            IRolePrivilegesChecker rolePrivilegesChecker)
         {
             _configValuesPersistentSource = configValuesPersistentSource;
+            _rolePrivilegesChecker = rolePrivilegesChecker;
         }
 
         /// <summary>
-        /// Get configuration values
+        /// Get a single non-secure configuration value by name. Restricted to administrators;
+        /// secure (secret) values are never returned through this endpoint.
         /// </summary>
         /// <param name="name">The name of the configuration value to retrieve</param>
         [HttpGet]
         public IActionResult GetConfigValue([FromQuery] string name)
         {
-            var value = _configValuesPersistentSource.GetConfigValue(name);
+            if (!_rolePrivilegesChecker.IsAdmin(User))
+            {
+                return StatusCode(StatusCodes.Status403Forbidden,
+                    "You are not authorized to read configuration values.");
+            }
+
+            var value = _configValuesPersistentSource.GetNonSecureConfigValue(name);
             if (string.IsNullOrWhiteSpace(value))
             {
                 return NotFound($"Config value '{name}' not found");
--- a/src/Dorc.Api/ConfigValuesController.cs
+++ b/src/Dorc.Api/ConfigValuesController.cs
@@ -1,4 +1,5 @@
 using Dorc.PersistentData;
+using Dorc.PersistentData.Sources;
 using Dorc.PersistentData.Sources.Interfaces;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Http;
@@ -23,20 +24,33 @@ namespace Dorc.Api.Controllers
         }
 
         /// <summary>
-        /// Get a single non-secure configuration value by name. Restricted to administrators;
-        /// secure (secret) values are never returned through this endpoint.
+        /// Get a single non-secure configuration value by name. Only administrators may read
+        /// configuration values through this endpoint, and secure (secret) values are never
+        /// returned here regardless of role.
         /// </summary>
         /// <param name="name">The name of the configuration value to retrieve</param>
         [HttpGet]
         public IActionResult GetConfigValue([FromQuery] string name)
         {
+            // The caller is authenticated (authorized) but only administrators have permission
+            // to read configuration values through this endpoint.
             if (!_rolePrivilegesChecker.IsAdmin(User))
             {
                 return StatusCode(StatusCodes.Status403Forbidden,
-                    "You are not authorized to read configuration values.");
+                    "You do not have permission to read configuration values; this requires the Admin role.");
             }
 
-            var value = _configValuesPersistentSource.GetNonSecureConfigValue(name);
+            string? value;
+            try
+            {
+                value = _configValuesPersistentSource.GetNonSecureConfigValue(name);
+            }
+            catch (SecureConfigValueRequestedException ex)
+            {
+                // Secure values are never exposed through this endpoint.
+                return StatusCode(StatusCodes.Status403Forbidden, ex.Message);
+            }
+
             if (string.IsNullOrWhiteSpace(value))
             {
                 return NotFound($"Config value '{name}' not found");
--- a/src/Dorc.Api/Controllers/BundledRequestsController.cs
+++ b/src/Dorc.Api/Controllers/BundledRequestsController.cs
@@ -28,6 +28,21 @@ namespace Dorc.Api.Controllers
         }
 
         /// <summary>
+        /// Returns a 403 result when the current user cannot modify the given project, or null
+        /// when the user is permitted. A missing project id is treated as not permitted.
+        /// </summary>
+        private IActionResult? EnsureCanModifyProject(long? projectId)
+        {
+            if (!projectId.HasValue ||
+                !_securityPrivilegesChecker.CanModifyProject(User, (int)projectId.Value))
+            {
+                return StatusCode(StatusCodes.Status403Forbidden, "User does not have Modify rights on this Project");
+            }
+
+            return null;
+        }
+
+        /// <summary>
         /// Get the list of request bundles
         /// </summary>
         /// <param name="projectNames"></param>
@@ -99,9 +114,10 @@ namespace Dorc.Api.Controllers
                 }
 
                 // Check user has write/modify rights for the project
-                if (!_securityPrivilegesChecker.CanModifyProject(User, (int)model.ProjectId.Value))
+                var forbidden = EnsureCanModifyProject(model.ProjectId);
+                if (forbidden != null)
                 {
-                    return StatusCode(StatusCodes.Status403Forbidden, "User does not have Modify rights on this Project");
+                    return forbidden;
                 }
 
                 _bundledRequestsPersistentSource.AddRequestToBundle(model);
@@ -136,9 +152,10 @@ namespace Dorc.Api.Controllers
                 }
 
                 // Check user has write/modify rights for the project
-                if (!_securityPrivilegesChecker.CanModifyProject(User, (int)model.ProjectId.Value))
+                var forbidden = EnsureCanModifyProject(model.ProjectId);
+                if (forbidden != null)
                 {
-                    return StatusCode(StatusCodes.Status403Forbidden, "User does not have Modify rights on this Project");
+                    return forbidden;
                 }
 
                 _bundledRequestsPersistentSource.UpdateRequestInBundle(model);
@@ -169,10 +186,10 @@ namespace Dorc.Api.Controllers
                 }
 
                 // Check user has write/modify rights for the project
-                if (!bundle.ProjectId.HasValue ||
-                    !_securityPrivilegesChecker.CanModifyProject(User, (int)bundle.ProjectId.Value))
+                var forbidden = EnsureCanModifyProject(bundle.ProjectId);
+                if (forbidden != null)
                 {
-                    return StatusCode(StatusCodes.Status403Forbidden, "User does not have Modify rights on this Project");
+                    return forbidden;
                 }
 
                 _bundledRequestsPersistentSource.DeleteRequestFromBundle(id);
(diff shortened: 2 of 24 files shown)


src/Dorc.Api/ConfigValuesController.cs
```diff
@@ -13,24 +12,36 @@ namespace Dorc.Api.Controllers
...
         /// </summary>
         /// <param name="name">The name of the configuration value to retrieve</param>
         [HttpGet]
         public IActionResult GetConfigValue([FromQuery] string name)
         {
-            var value = _configValuesPersistentSource.GetConfigValue(name);
+            if (!_rolePrivilegesChecker.IsAdmin(User))
+            {
+                return StatusCode(StatusCodes.Status403Forbidden,
+                    "You are not authorized to read configuration values.");
```
kirill-suspitsin-ext-sefe: Need to add information that only admins can get secure config values.
Also the message itself contains incorrect information, user is authorized but doesn't have enough permissions

src/Dorc.PersistentData/Sources/ConfigValuesPersistentSource.cs
```diff
@@ -30,6 +30,18 @@ public string GetConfigValue(string key, string defaultValue = null)
...
         }
 
+        public string? GetNonSecureConfigValue(string key)
+        {
+            using var context = _contextFactory.GetContext();
+            var configValue = context.ConfigValues.SingleOrDefault(x => x.Key == key);
+
+            // Never disclose secure values through this accessor.
+            if (configValue == null || configValue.Secure)
+                return null;
```
kirill-suspitsin-ext-sefe, line 39: Lets return an error here that secure variables has been requested in the non-secure logic

src/Dorc.Core/ApiCaller.cs
```diff
@@ -70,6 +70,56 @@ public ApiResult<T> Call<T>(Endpoints endpoint, Method method, Dictionary<string
             return result;
         }
 
+        // Property names, in priority order, that may carry a human-readable error message:
+        // "Message" (CopyEnvBuildResponseDto, serialized exceptions), "detail"/"title"
+        // (RFC 7807 ProblemDetails / ASP.NET validation responses).
+        private static readonly string[] ErrorMessagePropertyNames =
+            { "Message", "message", "detail", "title" };
```
kirill-suspitsin-ext-sefe, line 78: It looks ugly and in some cases error information can be passed in some different fields.
Lets maybe try to standardize our error messages with only one error field from all apis?

src/Dorc.Api/Controllers/BundledRequestsController.cs
```diff
@@ -99,9 +114,10 @@ public IActionResult Create([FromBody] BundledRequestsApiModel model)
                 }
 
                 // Check user has write/modify rights for the project
-                if (!_securityPrivilegesChecker.CanModifyProject(User, (int)model.ProjectId.Value))
+                var forbidden = EnsureCanModifyProject(model.ProjectId);
+                if (forbidden != null)
                 {
-                    return StatusCode(StatusCodes.Status403Forbidden, "User does not have Modify rights on this Project");
+                    return forbidden;
```
kirill-suspitsin-ext-sefe: Previous implementation looks better and more human-readable I think.
Maybe we just need to add need to add the "!projectId.HasValue" condition in if statement as it was earlier?

kirill-suspitsin-ext-sefe (changes requested):

src/Dorc.Core/ApiCaller.cs
```diff
@@ -70,6 +70,56 @@ public ApiResult<T> Call<T>(Endpoints endpoint, Method method, Dictionary<string
...
+                        if (root.TryGetProperty(propertyName, out var message)
+                            && message.ValueKind == JsonValueKind.String)
+                        {
+                            var value = message.GetString();
+                            if (!string.IsNullOrEmpty(value))
+                            {
+                                return value;
+                            }
+                        }
+                    }
```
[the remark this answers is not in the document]
benhegartysefe: Done in 56f56a9 — replaced the implicitly-filtering `foreach` with an explicit `Select` → `Where(ValueKind == String)` → `Select(GetString)` → `FirstOrDefault(non-empty)` pipeline, keeping the `responseContent` fallback unchanged.

---

src/Dorc.Api/ConfigValuesController.cs
benhegartysefe: Updated in 56f56a9. The 403 message now reads: *"You do not have permission to read configuration values; this requires the Admin role."* — reflecting that the caller is authenticated but lacks permission, rather than "not authorized".

Note on the design: secure values are never returned through this endpoint **for anyone** (requesting one now yields a distinct 403, see the related change on `ConfigValuesPersistentSource`), which mirrors the admin-only list endpoint that also doesn't decrypt secrets. If you'd instead prefer non-admins to be able to read *non-secure* values here (i.e. gate only secrets behind Admin), say the word and I'll adjust the gate.

---

src/Dorc.PersistentData/Sources/ConfigValuesPersistentSource.cs
benhegartysefe: Done in 56f56a9 — `GetNonSecureConfigValue` now throws `SecureConfigValueRequestedException` when a `Secure` key is requested through the non-secure accessor, instead of silently returning `null`. The controller catches it and returns a 403 with the explicit reason.

---

src/Dorc.Core/ApiCaller.cs
benhegartysefe: Agreed — probing multiple fields isn't ideal. For now I've cleaned it into a LINQ pipeline (56f56a9), but you're right that the proper fix is a single standardized error contract returned by every API (one `error` field). That's a cross-cutting change touching all controllers, and it dovetails with a separate concern about controllers returning raw exception objects to clients. I'd suggest doing it as a dedicated follow-up PR rather than widening this one; once the server side returns a single field, `ExtractErrorMessage` collapses to reading that one field. Happy to pick that up separately — want me to open it?

---

src/Dorc.Api/Controllers/BundledRequestsController.cs
benhegartysefe: Reverted in 56f56a9 — removed the `EnsureCanModifyProject` helper and restored the inline `CanModifyProject` checks. `Create`/`Update` keep the separate `!ProjectId.HasValue` → `BadRequest` guard followed by the inline 403, and `Delete` keeps the combined `!bundle.ProjectId.HasValue || !CanModifyProject(...)` guard.

---

src/Dorc.Core/ApiCaller.cs
kirill-suspitsin-ext-sefe: Yes, please. dedicated Pr would be good for this

src/Dorc.Api/ConfigValuesController.cs
kirill-suspitsin-ext-sefe: No, config values only available for admins even non-secure

benhegartysefe: Confirmed — the endpoint already requires Admin for **all** config values (secure and non-secure); non-admins get the 403 regardless. So the current implementation matches what you want. Thanks for clarifying.

---

src/Dorc.Core/ApiCaller.cs
benhegartysefe: Sounds good — I'll open a dedicated PR for a standardized single-field error contract across the APIs (and simplify `ExtractErrorMessage` to read that one field once it's in place). Tracking it separately so this PR stays scoped.

---

kirill-suspitsin-ext-sefe (approved):

PR #797, opened 2026-08-02
ci: build the solution in parallel and stop double-triggering runs

hegsie:
The GitHub Actions build was taking ~11m40s on main, with `Build solution`
alone accounting for 6m08s of it. Three flags on the msbuild invocation each
disabled a different form of concurrency:

  /m:1                            single MSBuild node
  /p:BuildInParallel=false        no parallel project batching
  /p:UseSharedCompilation=false   cold csc.exe start for every project

They arrived together in the commit that first added the workflow, with no
accompanying comment or failure they were working around. Setup.Dorc and
Setup.Acceptance declare all their ProjectReferences explicitly, so MSBuild's
dependency graph still orders the MSI harvest after the projects it packages
once parallel scheduling is enabled. Replaced with /m.

Also:

- Dropped topic branches from the push trigger. They are already covered by
  the pull_request trigger, and listing them in both meant every push to a
  branch with an open PR started two full ~12 minute Windows runs of the same
  commit.
- Added a concurrency group so a new commit supersedes an in-flight run.
  main and release/** are exempt, since those runs publish artifacts.
- Removed the `nuget restore` pass that ran immediately before
  `dotnet restore` over the same solution. Every project is SDK-style and
  there is no packages.config anywhere in the tree, so it was duplicated work.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01YY39MK8CWjmqWNcBjKzSD6


benhegartysefe (approved):

JamesShortSefe (approved):
looks sensible enough

PR #798, opened 2026-08-02
Responsive record views: GitHub-style narrow list rendering inside Vaadin Grid for all 37 grid views

hegsie:
# Responsive record views

Every grid view in `dorc-web` (all 37) now collapses to a **GitHub-mobile-style single-column list** at narrow container width, while desktop width is unchanged. One `<vaadin-grid>` element, two presentations — no second rendering engine, so Vaadin's virtualiser, `dataProvider` contract and grid keyboard/ARIA model apply at both widths.

## Problem (measured, not asserted)

The original hypothesis was "we've outgrown Vaadin". Investigation refuted that — DORC uses a conservative subset (0 tree grids, 0 selection models, 1 frozen column) — but found a real, measured defect underneath:

- **8 of 10 measured views required horizontal scrolling at 375px**, including both views on the on-call triage journey (`page-monitor-requests` +139px, `env-monitor` +79px; worst: `page-env-history` +389px with only 4 columns showing). Overflow does not track column count — a 3-column view with two hardcoded 300px widths was third-worst at +325px.
- The "cheap fix" (narrower explicit widths) was measured honestly: forcing the journey view's four columns into 375px **truncates ~54% of the identity field** — trading horizontal scrolling for losing exactly the content triage needs.
- The old mechanism (`ResponsiveMixin` + per-column `?hidden` bindings) was opt-in per file, so the worst views were the ones that never opted in, and "which fields matter on a phone" was undeclared across 180 column declarations in 37 files.

The measurement instrument is committed at `tests/measurements/u18-narrow-overflow.measure.ts` (deliberately outside the CI test glob; run on demand).

## Design

**Row templates.** Each view declares its field priority once, in a reviewable module under `src/row-templates/` — slots: **primary** (identity, full row width), **metadata** (muted secondary fields), **chip** (status, driven by the item model), **actions** (the view's existing row actions), **details** (disclosure carrying the fields that are hidden columns today). The disclosure affordance owns its `aria-expanded` — verified against Vaadin 25.2.6 source that row-details expose no ARIA state (that attribute is tree-grid-only, and DORC has no tree grids).

**Narrow-mode controller.** `NarrowListController` (a Lit `ReactiveController`) measures the host **container** via `ResizeObserver` — not the viewport — so dialog-hosted grids (`make-like-production`) collapse when the dialog is narrow. It replaces `ResponsiveMixin` in migrated views (the mixin now has zero grid-view consumers), ignores zero-width measurements (hidden hosts don't flap), and triggers `auto-width` recalculation on return to wide (Vaadin doesn't re-measure unhidden columns). Threshold: the existing shared 768px constant; derivation: the worst measured content demand was 764px.

**The header-hosted bar.** Filters, sorters and view-level controls (e.g. the live SignalR connection indicator and auto-refresh toggle that lived in a column header) move to a bar hosted in the list column's `headerRenderer`. This placement is load-bearing, proven by a committed spike test: hiding a column removes its cell content from the DOM, and disconnected sorters silently drop out of the grid's sort state (`_getActiveSorters()` filters on `isConnected`) — a floating bar above the grid would silently lose sort, including declarative defaults. Header-hosted sorters stay connected; the spike's control case demonstrates the failure mode it avoids.

**Activation semantics.** Where row activation previously opened the result page (`env-monitor`, and `page-monitor-requests`' id button), narrow-width activation now toggles disclosure and the open-result action lives on the id link in the primary line; wide-width behaviour is untouched.

**Batch reuse.** The 32 post-pilot views migrate via compact per-view templates (`batch-row-templates.ts`, one named export per view) over shared helpers — notably `hostCell`, which reuses each view's existing imperative cell renderers inside list slots so action buttons/links keep their exact behaviour (with a stub carrying the host under every `_column.*` property name the renderers dereference).

## Screenshot-verified polish rounds

Every migrated view was captured at 375px **with representative data in every row** by a second committed instrument, `tests/measurements/screenshot-all-grids.measure.ts` (36 captures; endpoints served by an XHR shim so property-driven and dataProvider-driven views alike render real content — a capture of an empty grid verifies nothing). Eyes-on review of those captures drove several fixes:

- **Access Control dialog**: migrated from `paper-dialog` to the project's `hegs-dialog` and sized for phones (`min(84vw, 620px)` body, viewport-capped height).
- **Duplicate tag registration**: two components both registered `<env-deployments>`; the orphaned card variant was unused and has been deleted (suite drops from 125 to 123 tests with it).
- **`component-previous-attempts`**: its wide log renderer crushed the narrow row to one character per line; the row template was rewritten with a compact log action.
- **Form-control alignment** (latest round): narrow captures showed a mishmash of per-field fixed widths — 6em/7.5em filter fields in the monitor bars, 400px-minimum combos inside `<table>`-cell layouts, toolbars pairing 50%-wide search boxes with 250px buttons. Individually sized for a desktop row, they misaligned badly once stacked at phone width. Replaced with a few shared flex rules in `dorc-list-row.ts` so every row sizes fields the same way: bar fields flex equally (`flex: 1 1 8rem`); a shared `.dorc-toolbar` class (search flexes full-width, buttons keep natural width, the row wraps) adopted by the seven list-page toolbars, with `page-environments-list` dropping its fixed-position header for normal flow; the table-cell form rows in `add-edit-access-control`, `page-variables` and `env-variables` became flex rows with the fixed minimums removed; and list rows now wrap wide action groups below the identity text instead of crushing it (small icon actions are unaffected).

The sweep itself was also hardened to the "no data, not checked" standard: list pages' connect-time fetches were silently wiping the injected rows after setup (those captures were empty in the baseline too), and `env-variables` had never rendered at all behind its `envLoaded` gate — both now inject after the view settles, which is what surfaced the wide-actions issue above.

## Provenance

The design went through a documented adversarial process (working artefacts removed in the final commit; summarised here): three review rounds on the problem statement (which caught a ~2.5× census-counting error, the false "3-column ceiling" claim, the sorter-disconnection design bug, and a fabricated measurement transcription), two rounds on the implementation sequence, an in-browser measurement pass, and a piloted delivery (5 views) before the batch (32 views). axe-core was added as a devDependency under explicit approval to give the narrow rendering automated a11y assertions.

## Testing

- Full suite **123/123 green**; `tsc`, test typecheck and eslint clean. (Chromium locally; CI runs chromium/firefox/webkit.)
- New coverage: controller container-tracking (including narrow-container-in-wide-window and zero-width cases), the bar-mechanism spike (sort connectivity both ways), container-driven collapse + restore (SC10 crossing, incl. `auto-width` recalc) on migrated components, axe a11y suite over the narrow rendering, and actions-reachability assertions.
- Idiom-asserting tests that encoded the old `matchMedia`/`?hidden` mechanism were replaced with container-driven equivalents; behaviour-asserting tests pass unmodified.
- All 36 narrow-width captures re-generated and reviewed after the polish rounds.

## Known follow-ups (deliberate, recorded)

1. **Human validation**: the scripted phone walkthrough by a non-author and the design side-by-side against GitHub's mobile list have not been executed — the built-for personas (app support / on-call) have no current usage or telemetry, so a session with a real on-call engineer is the committed next step, and the batch views' field priorities (authored from column specs) should be confirmed there too.
2. Batch DP views' narrow bars carry **sorters only**; their filter inputs remain wide-width-only (the pilot's journey views have full filter bars).
3. `hostCell`'s detached-span reuse is behaviour-preserving; the clean long-term form is extracting those renderers into templates (done for `page-permissions-list` as the reference).
4. 26 files still hardcode the 768px value in CSS media queries; the JS threshold lives in one module.
5. New-view rule now in force: grid-bearing views ship a row-template module and use the controller; `?hidden` column bindings should be rejected in review.
6. `page-servers-list` / `page-databases-list` chip renderers reference `--dorc-*` theme variables without Lumo fallbacks (invisible outside the app theme; suggested form `var(--dorc-chip-bg, var(--lumo-contrast-10pct))`), and three other components still use `paper-dialog`.

https://claude.ai/code/session_013ginU1TnGCQ6WmZtjXucKD

---
_Generated by [Claude Code](https://claude.ai/code/session_013ginU1TnGCQ6WmZtjXucKD)_


hegsie:
## Triage outcome — replacement/rebase required

The narrow-list presentation itself is **not already in `main`**: `dorc-list-row.ts` and `narrow-list-controller.ts` are absent there. However, the earlier responsive work from PR #716 is already merged (`c8ac5e7e`), so current `main` already provides the existing mobile column-hiding baseline.

I fetched current `main` and attempted a normal `git merge origin/main --no-edit` in a dedicated PR #798 worktree. It produces **34 content conflicts**, concentrated in the same 34 web components/pages that this PR migrates. This is not safe to resolve mechanically: the PR predates and does not contain the Vaadin 25/dialog/router migration in `f9008e98` (#799), while both changes extensively rewrite the affected components. Choosing either side would silently discard either current platform migration work or the responsive-list behavior.

The sole automated review thread is already resolved, and the PR's last recorded checks passed only against its August base (`40f106a`), not current `main`.

**Recommendation:** close this stale branch rather than force a conflict-heavy rewrite into it. If the GitHub-style narrow-list UX remains a product priority, open a newly scoped successor from current `main`, port each view incrementally with fresh tests and visual validation, and retain this PR as the design/reference history.

PR #828, opened 2026-08-14
S-002/S-003: authorize daemon status reads; fix combined-mode scope policy

hegsie:
Closes #827

**Stacked on #826.** Merge that first.

## Daemon status

Both `Get` overloads now resolve the environment and require `CanModifyEnvironment` **before** invoking the probe. The probe performs a `LogonUser` with the deployment credential, so the ordering was the whole weakness — the side effect happened, then the caller was considered.

## Scope policy

The requirement now registers for `OAuth` **and** `Both`. The estate runs `OAuth+WinAuth`, so it had never applied.

The requirement is scoped to bearer requests via `RequireAssertion`: a Windows-authenticated caller has no `scope` claim and must not be expected to have one, so applying it unconditionally in combined mode would lock out every Windows user. That assertion is the part worth reviewing.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KT2gJnb8LC4DGHT2bDTCJk)_


Otar-Khokhashvili-ext-sefe (approved):

src/Dorc.Api/Controllers/DaemonStatusController.cs
```diff
@@ -35,11 +35,24 @@ public DaemonStatusController(IDaemonStatusProbe daemonStatusProbe, ISecurityPri
...
                 return StatusCode(StatusCodes.Status400BadRequest, "Environment ID is not valid!");
 
+            var environment = _environmentsPersistentSource.GetEnvironment(id, User);
+            if (environment == null)
+                return NotFound($"Environment with ID {id} not found");
+
+            // Probing daemon statuses logs on with the deployment credential for the target
+            // environment's tier, so authority must be established before the probe is
+            // invoked - not merely reflected in what it returns.
+            if (!_securityPrivilegesChecker.CanModifyEnvironment(User, environment.EnvironmentName))
```
hegsie: This is only a GET method, it should be not be blocking reads of daemons statuses

src/Dorc.Api/Program.cs
```diff
@@ -438,9 +459,13 @@ static void AddSwaggerGen(IServiceCollection services, IConfigurationSettings co
 app.UseMiddleware<WinAuthLoggingMiddleware>();
 
 var endpointConventionBuilder = app.MapControllers();
-if (authenticationScheme is ConfigAuthScheme.OAuth)
+// Enforce Authorization Policy [see constant 'apiScopeAuthorizationPolicy'] on all the
+// Controllers. This must cover ConfigAuthScheme.Both as well as OAuth: Both is the distinct
+// string "OAuth+WinAuth", so an exact match against OAuth alone silently skipped the policy
+// and left bearer tokens unchecked for scope in combined mode. The policy itself exempts
+// Windows-authenticated requests, so applying it in Both mode does not deny them.
+if (authenticationScheme is ConfigAuthScheme.OAuth or ConfigAuthScheme.Both)
```
benhegartysefe: WinAuth was deliberatly disabled in preperation for complete removal, why are we adding it back in?

src/Dorc.Api/Controllers/DaemonStatusController.cs
hegsie: Addressed in 87e11540. I removed the environment write-permission checks from both daemon-status GET overloads and deleted the associated authorization tests, so authenticated users retain the existing read behaviour. The security plan now records S-002 as declined and leaves the underlying credential-resolution risk for S-021.

src/Dorc.Api/Program.cs
hegsie: This change does not add or enable WinAuth. `ConfigAuthScheme.Both` and its registration already exist; the change only attaches the scope policy when that existing mode is selected, and the policy checks scope only for bearer requests. I clarified that explicitly in Program.cs and narrowed the PR title/body to S-003.

benhegartysefe: then why do you need both?

src/Dorc.Api/Program.cs
benhegartysefe: ConfigureOAuth should be the default now, and winAuth & Both shouldn't be supported any longer

src/Dorc.Api/Program.cs
hegsie: Follow-up with the repository history: WinAuth was removed in ba1ff614 but restored in 0cf4dfc1, which is on current main; OAuth+WinAuth also remains the default in appsettings and the installer template. This PR does not add or enable it—it closes the bearer-scope gap while combined mode still exists. #713 is the separate planned removal. Commit 1b4db617 now makes that transitional relationship explicit in the security plan.

src/Dorc.Api/Controllers/DaemonStatusController.cs
hegsie: Follow-up on the existing response: confirmed on the current head that both daemon-status GET overloads retain their original authenticated read behaviour and no environment write-permission check remains. The plan records S-002 as declined and leaves credential-resolution removal to S-021.

kirill-suspitsin-ext-sefe (approved):

kirill-suspitsin-ext-sefe (approved):

PR #852, opened 2026-08-17
Fix cross-node in-flight request cancellation

### LINKED ISSUE #818: Cancellation job
As for now, we have 2 DOrc Monitor nodes that are working in parallel.
If the first node has started some deployment request processing and user cancel this request, second node can process this cancellation. In this case first node doesn't now about this cancellation and will continue to work on the current component in the cancelled request that can take a lot of time.
Need to create some background job that will monitor cancellation/restarting of the deployments and kill the current runner process for these requests.

Otar-Khokhashvili-ext-sefe:


src/Dorc.Monitor/RequestStatusPoller.cs
```diff
@@ -0,0 +1,132 @@
...
+    /// <summary>
+    /// Terminal states that should trigger cancellation of an in-flight request.
+    /// If a request transitions to any of these states while we're processing it,
+    /// we must abort immediately.
+    /// </summary>
+    private static readonly HashSet<string> TerminalCancelStates = new(StringComparer.OrdinalIgnoreCase)
+    {
+        nameof(DeploymentRequestStatus.Cancelled),
+        nameof(DeploymentRequestStatus.Abandoned),
+        nameof(DeploymentRequestStatus.Cancelling),
```
kirill-suspitsin-ext-sefe: I think Cancelling status here can be removed. This status could be set up only by webapi not by the another node of monitor. And cancellation process can be executed by the same node that is working on the request, such situation can lead to warning and error.
Also need to think how we can handle the same problem for restarting of the request. When one node is executing the request and the another one will restart this.

kirill-suspitsin-ext-sefe (approved):

